PR #15, opened 2022-11-18
Add notification panel

LomyW:
**User can display dummy notifications on a panel by click on a bell button.**
as discussed: NotificationPanel with test notification types.


src/components/Notification/MessageBox.tsx
```diff
@@ -0,0 +1,34 @@
+import { Box, HStack, Spacer, Text, VStack } from 'native-base'
+import { getIconForCategory } from '../../helper/notification-helper'
+import { UserNotification } from '../../types/lernfair/Notification'
+import TimeAgo from './TimeAgo'
+
+type Props = {
+  userNotification: UserNotification
+  isToast?: boolean
```
jeffersondarcy, line 8: maybe it's better to call it `displayTime` or `showTime` to make it more universal

src/components/Notification/MessageBox.tsx
```diff
@@ -0,0 +1,34 @@
...
+}
+
+const MessageBox: React.FC<Props> = ({ userNotification, isToast, isRead }) => {
+  const { headline, body, notification, createdAt } = userNotification
+
+  return (
+    <Box borderRadius={2} bgColor={isRead ? '' : 'primary.100'} mb={2} h="54px">
+      <HStack alignItems="center" space={1}>
+        <VStack>
+          <Box p={'8px'}>{getIconForCategory(notification.messageType)}</Box>
```
jeffersondarcy, line 19: the naming looks a bit inconsistent with mixing of type and category, maybe it's better to rename the function to just `getIcon`

src/components/Notification/NotificationAlert.tsx
```diff
@@ -0,0 +1,41 @@
+import { Button, Text, Circle, Popover, VStack } from 'native-base'
+import BellIcon from '../../assets/icons/lernfair/lf-bell.svg'
+import NotificationPanel from './NotificationPanel'
+
+const NotificationAlert: React.FC = () => {
+  // TODO: the implementation is done in the upcoming PR
+  const count: number = 4
+
+  return (
+    <>
```
jeffersondarcy, line 10: you don't need the Fragment <> because everything is wrapped inside Popover anyway

src/components/Notification/NotificationAlert.tsx
```diff
@@ -0,0 +1,41 @@
...
+                  alignSelf="flex-start"
+                  bgColor="danger.500"
+                  size="3.5"
+                  zIndex={1}>
+                  <Text fontSize="10" color="white">
+                    {count}
+                  </Text>
+                </Circle>
+              )}
+              <Button {...triggerProps} bgColor="">
```
jeffersondarcy, line 30: props spreading should be avoided if possible

src/components/Notification/NotificationAlert.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const NotificationAlert: React.FC = () => {
+  // TODO: the implementation is done in the upcoming PR
+  const count: number = 4
+
+  return (
+    <>
+      <Popover
+        placement="bottom"
+        trigger={triggerProps => {
```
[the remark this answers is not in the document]
LomyW: its from native base. But yes it is possible to trigger a separated component.

src/components/Notification/MessageBox.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { UserNotification } from '../../types/lernfair/Notification'
+import TimeAgo from './TimeAgo'
+
+type Props = {
+  userNotification: UserNotification
+  isToast?: boolean
+  isRead?: boolean
+}
+
+const MessageBox: React.FC<Props> = ({ userNotification, isToast, isRead }) => {
```
jeffersondarcy, line 12: should isToast and isRead have explicit default values?

LomyW: why?

src/components/Notification/NotificationPanel.tsx
```diff
@@ -0,0 +1,65 @@
...
+          <Popover.Header>
+            <Box alignSelf="flex-end" mr={10}>
+              <SettingsIcon />
+            </Box>
+          </Popover.Header>
+          <Popover.Body>
+            <Box w="320" maxH="580">
+              <ScrollView>
+                <Box w="320">
+                  {data.me.concreteNotifications.map(
```
jeffersondarcy, line 40: it woul look nicer ond hide implementation details if you could just do `concreteNotifications.map` without `data.me`

src/hooks/useNotificationPanel.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const minutesDiff = now.diff(createdAt, 'minutes').minutes
+  const hoursDiff = now.diff(createdAt, 'hours').hours
+  const daysDiff = now.diff(createdAt, 'days').days
+  const weeksDiff = now.diff(createdAt, 'weeks').weeks
+  const timeDiffString = timeDiff.toFormat('m')
+
+  return { minutesDiff, hoursDiff, daysDiff, weeksDiff, timeDiffString }
+}
+
+const useTimeDifference = (timestamp: string, isToast?: boolean) => {
```
jeffersondarcy, line 40: this looks for me like a helper function, not a hook

src/hooks/useNotificationPanel.tsx
```diff
@@ -0,0 +1,68 @@
+import { gql, useQuery } from '@apollo/client'
```
jeffersondarcy, line 1: I think the file name useNotificationPanel.tsx is misleading as there is no hook with such name, also the extention should be "ts" as there is no JSX

src/hooks/useNotificationPanel.tsx
LomyW: its a hook because of the useTranslation hook

src/hooks/useNotificationPanel.tsx
LomyW: renamed it to useNotifications.ts

src/hooks/useNotificationPanel.tsx
LomyW: added getTimeDifference to the notification-helper

LomyW: renamed it to useTimeAgoTranslation

src/components/Notification/MessageBox.tsx
```diff
@@ -0,0 +1,38 @@
...
+          <Box p={'8px'}>{getIcon(notification.messageType)}</Box>
+        </VStack>
+        <VStack>
+          <Text bold fontSize="14px" mt={2}>
+            {headline}
+          </Text>
+          <Text fontSize="12px">{body.slice(0, 30)}</Text>
+        </VStack>
+        <Spacer />
+        <Box>{!displayTime && <TimeIndicator createdAt={createdAt} />}</Box>
```
timidak, line 32: Why are we rendering the `<TimeIndicator />` when displayTime is false?
I would suggest to do it the other way around

src/components/Notification/NotificationAlert.tsx
```diff
@@ -0,0 +1,45 @@
...
+        {count && (
+          <Circle
+            //_web
+            position="absolute"
+            my={3}
+            alignSelf="flex-start"
+            bgColor="danger.500"
+            size="3.5"
+            zIndex={1}>
+            <Text fontSize="10" color="white">
```
timidak, line 26: Can we use xs, md or other standardized values for fontSize?

src/components/Notification/NotificationAlert.tsx
```diff
@@ -0,0 +1,45 @@
...
+              {count}
+            </Text>
+          </Circle>
+        )}
+        <Button onPress={onPress} ref={ref} bgColor="">
```
timidak, line 31: remove the `bgColor` prop if it's empty

src/components/Notification/TimeIndicator.tsx
```diff
@@ -0,0 +1,33 @@
...
+  }, 60_000)
+
+  return (
+    <Box pr={3} mt={-3}>
+      {typeof time === 'string' ? (
+        <Text fontSize="12px">{time}</Text>
+      ) : (
+        <Text fontSize="12px">{t(time.text, time?.options)}</Text>
+      )}
+      {/* {time && <Text fontSize="12px">{t(time.text, time?.options)}</Text>} */}
```
timidak, line 28: Can this be removed? 🙂

src/components/SettingsButton.tsx
```diff
@@ -1,15 +1,14 @@
...
+import { Button, HamburgerIcon } from 'native-base'
 import { useNavigate } from 'react-router-dom'
 
 type Props = {}
 
 const SettingsButton: React.FC<Props> = () => {
   const navigate = useNavigate()
   return (
-    <TouchableOpacity onPress={() => navigate('/settings')}>
+    <Button bgColor="" ml={-4} onPress={() => navigate('/settings')}>
```
timidak, line 9: Can we remove the background color prop? And is it possible to use the margin on the other side, rather than using a negative margin?

src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,27 @@
...
+        notification {
+          messageType
+        }
+        createdAt
+      }
+    }
+  }
+`
+
+const useNotifications = (skip?: number, take?: number) => {
```
timidak, line 20: I don't think we need `skip?` and `take?`, maybe we'll need it later, but I would remove it for now

src/hooks/useNotifications.ts
```diff
@@ -0,0 +1,27 @@
...
+  const { data, loading, error, refetch } = useQuery(userNotificationQuery)
+  // TODO: on first request me seems to be empty, not evaluation
+  const notifications = data?.me?.concreteNotifications
+  return { data, notifications, loading, error, refetch }
```
timidak, line 24: You should either use an effect here or do not destructure the notifications. Otherwise the notifications are not updated properly.
In addition, I wouldn't just return data, but only the notifications.
```suggestion
  const { data, loading, error, refetch } = useQuery(userNotificationQuery)
  
  const [notifications, setNotifications] = useState([]);
  
  useEffect(() => {
    if(!loading && !error) {
      setNotifications(data?.me?.concreteNotifications);
    }
  }, [loading]);
  
  return { notifications, loading, error, refetch }
```

timidak (changes requested):
Some minor changes left, but nice work 🎉 
One point I also saw is that we have no margin or padding to the edge of the browser, when opening the notifications panel. I guess this should be tackled as well.

PR #22, opened 2022-11-21
Websocket Client + Context

jeffersondarcy:
-WebSocket Client and Notifications Context implemented

tested with modified backend:

create new concrete notification as admin  in apollo playground (add Header
`{ "authorization": "Basic YWRtaW46SFo5NTU3Qk1HUEZRMlBWWjY5Q05MVkZCQjJUQTRIUEY=" }`
)
https://typedigitalworkspace.slack.com/archives/C049KSP9551/p1669131058081499

```
mutation { 
	pupilCreateMatchRequest(pupilId: 1)
}
mutation { 
	studentCreateMatchRequest(studentId: 2)
}
mutation { 
	matchAdd(poolName: "lern-fair-now" pupilId: 1 studentId: 2)
}
```

add at line 40 in common/websocket/index.ts on backend for testing:
```
        setInterval(async () => {
            const userId = this.instance.connectedUsers.keys().next().value;
            if (userId) {
                await this.instance.sendMessageToUser('pupil/1', { concreteNotificationId: 4 });
                log.debug('call send');
            }
        }, 3000);
```
this triggers only one notification to pop-up as the concreteNotificationId doesn't change


src/components/WebsocketClient.tsx
```diff
@@ -0,0 +1,39 @@
+import { useEffect, useRef, useContext} from "react"
+import { NotificationsContext } from "../hooks/NotificationsProvider"
+
+export const WebsocketClient = () => {
```
timidak, line 4: I provided a first draft to @LomyW with an interface, a ws client implementation and a possible switch for a React Native implementation. Why aren't you using this?

src/hooks/NotificationsProvider.tsx
```diff
@@ -0,0 +1,26 @@
+import { createContext, FC, ReactNode, useState } from "react"
+import { UserNotification } from "../types/lernfair/Notification"
+
+type Notification = Partial<UserNotification>
```
timidak, line 4: Why a Partial?

src/components/WebsocketClient.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const {dispatchNewNotificationIds} = useContext(NotificationsContext)
+  const ws = useRef<WebSocket | null>(null);
+  
+
+  useEffect(() => {
+    // only for testing
+    const socket = new WebSocket("wss://ws.kraken.com");
```
timidak, line 11: Keep in mind that the connection can only be done when the user is logged in. Because you'll need the `Session token` to authenticate at the websocket server.

src/hooks/NotificationsProvider.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+type NotificationContextValue = {notifications: Notification[], dispatchNewNotificationIds: Function};
+export const NotificationsContext = createContext<NotificationContextValue>({notifications: [], dispatchNewNotificationIds: () => null})
+
+type State = { notifications: Notification[] }
+
+export const NotificationsProvider: FC<{ children: ReactNode }> =
+  ({ children })=> {
+    const [{notifications}, setState] = useState<State>({notifications: []})
```
timidak, line 13: What about: 
```suggestion
    const [notifications, setState] = useState<State>([])
```

The state would then be `Notification[]`

src/App.tsx
```diff
@@ -10,21 +10,27 @@ import './web/scss/index.scss'
 import FullPageModal from './widgets/FullPageModal'
 import { LFModalProvider } from './hooks/useModal'
 import { LernfairProvider } from './hooks/useLernfair'
+import { NotificationsProvider } from './hooks/NotificationsProvider'
+import { WebsocketClient } from "./components/WebsocketClient"
 
-function App() {
+function App () {
```
timidak, line 16: Where does this whitespace come from? Some formatter?

src/hooks/NotificationsProvider.tsx
```diff
@@ -0,0 +1,26 @@
...
+    const dispatchNewNotificationIds = async (notificationIds: Array<string | number>) => {
+      // fetch notification from graphql
```
timidak, line 15: I don't think you can fetch the notification here, if you use a query hook of Apollo.

timidak (changes requested):
Some changes, just ping me when the PR is ready for review. 🙂

src/hooks/NotificationsProvider.tsx
jeffersondarcy: that was only for testing purposes, will change it to "full"

src/hooks/NotificationsProvider.tsx
jeffersondarcy: I was thinking about mowing it to WS Client Component - one more reason for this

src/components/WebsocketClient.tsx
jeffersondarcy: will implement this logic later when dealing with real backend

src/App.tsx
jeffersondarcy: fixed

src/hooks/NotificationsProvider.tsx
```diff
@@ -0,0 +1,36 @@
+import { createContext, FC, ReactNode, useState } from "react"
+import { UserNotification } from "../types/lernfair/Notification"
+
+type Notification = UserNotification
+type State = { notifications: Notification[], notificationIds: number[] }
```
timidak, line 5: Why do we need the notifications and the ids. Aren't the ids enough?

src/hooks/NotificationsProvider.tsx
```diff
@@ -0,0 +1,36 @@
...
+export const NotificationsProvider: FC<{ children: ReactNode }> =
+  ({ children }) => {
+    const [state, setState] = useState<State>(
+      { notifications: [], notificationIds: [] })
+    const setNotificationIds = (notificationIds: number[]) => {
+      setState({ ...state, notificationIds })
+    }
+    const setNotifications = (notifications: Notification[]) => {
+      setState({ ...state, notifications })
+    }
```
timidak, line 23: This feels a little bit like React 15 for me, why aren't we using two separate `useState` functions for this?

timidak (changes requested):
Some changes to be made

src/hooks/NotificationsProvider.tsx
jeffersondarcy: first websocket sets the id, than the data component fetches the message(notification). when message updates the toast appears. notification id change triggers data fetching

src/hooks/NotificationsProvider.tsx
jeffersondarcy: true, I will change this

src/components/WebsocketClient.tsx
```diff
@@ -0,0 +1,48 @@
...
+    if (sessionState !== 'logged-in') {
+      wsClient.close()
+      
+      return
+    }
+    
+    // only for testing
+    //wsClient.connect("wss://ws.kraken.com");
+    
+    const host = 'wss://ws.localhost:5000'
```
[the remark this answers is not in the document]
jeffersondarcy: added a separate env param

src/hooks/NotificationsProvider.tsx
jeffersondarcy: done

src/components/NotificationsData.tsx
```diff
@@ -0,0 +1,13 @@
...
+export const NotificationsData: FC<{}> = () => {
+  const {notificationIds, setNotifications} = useContext(NotificationsContext)
+
+  useEffect(() => {
+    // fetch notifications from backend
+    // setNotifications(fetchedData)
+  }, [notificationIds])
+  
+  return null
+}
```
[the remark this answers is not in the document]
jeffersondarcy: I think in future the data will be used by notification panel too

src/types/Websocket.ts
```diff
@@ -0,0 +1,48 @@
+export type WebSocketClient = {
+  isConnected: boolean
+  connect: (url: string) => void
+  onMessage: (callback: (this: WebSocket, ev: MessageEvent) => any) => void
+  close: () => void
+}
+
+class BrowserWebsocketClient implements WebSocketClient {
```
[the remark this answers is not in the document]
jeffersondarcy: moved to lib

.env.example
```diff
@@ -0,0 +1,8 @@
+REACT_APP_BACKEND_URL="http://localhost:5000/apollo"
+REACT_APP_WEBSOCKET_URL="ws://localhost:5000"
+
+REACT_APP_MATOMO_URL="https://lernfair.matomo.cloud/"
+REACT_APP_DISABLE_MATOMO="true"
+
+REACT_APP_UNSPLASH=""
```
Jonasdoubleyou, line 7: Document in README (`cp .env.example .env`), also the default dev workflow should point to the development backend (as the majority does not have a local backend running). Can both be done somewhen ...

src/components/NotificationsProvider.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  return true
+}
+
+export const NotificationsProvider: FC<{ children: ReactNode }> =
+  ({ children }) => {
+    const [concreteNotificationId, setConcreteNotificationId] = useState<number>()
+    const [message, setMessage] = useState<UserNotification>()
+
+    const setValidatedMessage = (message: UserNotification | unknown) => {
```
Jonasdoubleyou, line 31: `useCallback`, otherwise the ContextProvider rerenders pretty often, no?

src/components/ToastNotifications.tsx
```diff
@@ -0,0 +1,19 @@
...
+export const ToastNotifications: FC = () => {
+  const {message} = useContext(NotificationsContext)
+  
+  const isMobile = useBreakpointValue({
+    base: true,
+    lg: false
+  })
+
+  useEffect(() => {
+    if (typeof message === 'object') showInAppMessage(message, isMobile)
```
Jonasdoubleyou, line 15: if (message != null) is clearer IMO

src/lib/Websocket.ts
```diff
@@ -0,0 +1,41 @@
...
+    this.socket.addEventListener('message', callback)
+  }
+  
+  close() {
+    this.socket.close()
+  }
+}
+
+// Here the switch could be done between websocket or the react native implementation
+const WSClient = BrowserWebsocketClient
```
Jonasdoubleyou, line 39: Nice!

src/components/NotificationsData.tsx
```diff
@@ -0,0 +1,17 @@
...
+  const {data, error} = useConcreteNotification(concreteNotificationId)
+
+  useEffect(() => {
+    if (!error && data) {
+      setMessage(data.concrete_notification)
+    }
+  }, [data, error])
+  
+  return null
+}
```
timidak, line 17: Do we need this in a separate Component. How about adding this to the ToastNotifications?

src/components/NotificationsData.tsx
```diff
@@ -0,0 +1,17 @@
...
+  
+export const NotificationsData: FC = () => {
+  const {concreteNotificationId, setMessage} = useContext(NotificationsContext)
+  const {data, error} = useConcreteNotification(concreteNotificationId)
+
+  useEffect(() => {
+    if (!error && data) {
+      setMessage(data.concrete_notification)
+    }
+  }, [data, error])
```
timidak, line 14: Adding the loading state of apollo here might be easier and less expensive I'd say.

src/components/WebsocketClient.tsx
```diff
@@ -0,0 +1,38 @@
...
+    }
+    if (!wsClient.current) {
+      const propagateMessage = ({ data }: MessageEvent) => {
+        const id = JSON.parse(data).concreteNotificationId
+        if (typeof id === "number") {
+          setConcreteNotificationId(id)
+        }
+      }
+      
+      const url = encodeURI(`${process.env.REACT_APP_WEBSOCKET_URL}?id=${userId}&token=${getSessionToken()}`)
```
timidak, line 29: I would add the WEBSOCKET url to the config.ts and use it from there imho, this makes the dependency clearer

src/hooks/useConcreteNotification.ts
```diff
@@ -15,11 +15,12 @@ const concreteNotificationQuery = gql`
...
 `
 
-const useConcreteNotification = (id: number) => {
+const useConcreteNotification = (id?: number) => {
   const { data, loading, error, refetch } = useQuery(
     concreteNotificationQuery,
     {
-      variables: { id }
+      variables: { id },
+      skip: typeof id !== 'number'
```
timidak, line 23: You're skipping here! Nice!

src/components/WebsocketClient.tsx
```diff
@@ -0,0 +1,38 @@
...
+      wsClient.current = new WSClient(url);
+      wsClient.current.onMessage(propagateMessage);
+      
+    }
+    return close
+  }, [sessionState, userId, setConcreteNotificationId]);
```
timidak, line 35: You'll need to wrap the `setConcreteNotificationId` in a useCallback in the Provider so that re-running the effect is avoided because of reference changes.

timidak (changes requested):
Some other changes and suggestions 😄

src/components/WebsocketClient.tsx
Jonasdoubleyou: Good point, also this is actually needed as we build the frontend only once (where REACT_APP_ is inlined into the bundle) and then use it both for staging & production. Please also introduce RUNTIME_WEBSOCKET_URL (like RUNTIME_BACKEND_URL)

.env.example
jeffersondarcy: removed the values and renamed to .env.template. also added a sentence in readme

src/components/WebsocketClient.tsx
jeffersondarcy: moved to hook

src/components/WebsocketClient.tsx
jeffersondarcy: done

src/components/NotificationsData.tsx
jeffersondarcy: loading is checked in useConcreteNotification.ts now

src/components/NotificationsData.tsx
jeffersondarcy: fixed

src/components/NotificationsProvider.tsx
jeffersondarcy: useState is removed now

src/hooks/useConcreteNotification.ts
```diff
@@ -15,22 +16,34 @@ const concreteNotificationQuery = gql`
...
+  return true
+}
+
+export const useConcreteNotification = (id: number | null) => {
+  const { data, loading, error } = useQuery(
     concreteNotificationQuery,
     {
-      variables: { id }
+      variables: { id },
+      skip: id == null
```
timidak, line 37:
```suggestion
      skip: !id
```
This should handle undefined, null & empty string. 🙂

src/hooks/useConcreteNotification.ts
```diff
@@ -15,22 +16,34 @@ const concreteNotificationQuery = gql`
...
     }
   )
-  const [notification, setNotification] = useState()
+  const [concreteNotification, setConcreteNotification] = useState(null)
 
   useEffect(() => {
-    if (!loading && !error) {
-      setNotification(data?.concrete_notification)
+    if (!loading && !error && data && isMessageValid(data?.concrete_notification)) {
```
timidak, line 43: I assume isMessageValid should return invalid if data is nullish so the third element is unnecessary, i guess.

src/hooks/useConcreteNotification.ts
```diff
@@ -15,22 +16,34 @@ const concreteNotificationQuery = gql`
   }
 `
 
-const useConcreteNotification = (id: number) => {
-  const { data, loading, error, refetch } = useQuery(
+const isMessageValid = (message: UserNotification | null): boolean => {
+  if (message === null) return false
```
timidak, line 20: Can it also be undefined?

timidak (commented):
Some last questions and suggestions 👍🏼

src/components/ToastNotifications.tsx
```diff
@@ -0,0 +1,21 @@
+import { useEffect, FC } from "react"
+import { useBreakpointValue } from "native-base"
+import { showInAppMessage } from "../widgets/InAppMessage"
+import { useConcreteNotificationMessage } from "../hooks/useConcreteNotificationMessage"
+import { useIncomingWSConcreteNotificationId } from "./useIncomingWSConcreteNotificationId"
+
+export const ToastNotifications: FC = () => {
+  const concreteNotificationId = useIncomingWSConcreteNotificationId()
+  const message = useConcreteNotificationMessage(concreteNotificationId)
```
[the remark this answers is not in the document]
jeffersondarcy: useConcreteNotification hook returns only message data now

src/components/useIncomingWSConcreteNotificationId.ts
```diff
@@ -0,0 +1,36 @@
...
+    wsClient.current = null
+  }
+  
+  useEffect(() => {
+    if (sessionState !== 'logged-in' || !userId) {
+      close()
+      return
+    }
+    if (!wsClient.current) {      
+      const url = encodeURI(`${process.env.REACT_APP_WEBSOCKET_URL}?id=${userId}&token=${getSessionToken()}`)
```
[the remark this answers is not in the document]
jeffersondarcy: this was the right file, but now I renamed it

src/components/useIncomingWSConcreteNotificationId.ts
```diff
@@ -0,0 +1,36 @@
+import { useEffect, useRef, useState } from "react"
+import { WSClient, WebSocketClient } from '../lib/Websocket'
+import { getSessionToken, useUserAuth } from "../hooks/useApollo"
+
+type State = number | null
```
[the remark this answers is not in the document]
jeffersondarcy: fixed

src/hooks/useConcreteNotificationMessage.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const useConcreteNotificationMessage = (concreteNotificationId: number | null): UserNotification | null => {
+  const {data, error} = useConcreteNotification(concreteNotificationId)
+  const [message, setMessage] = useState<UserNotification | null>(null)
+
+  useEffect(() => {
+    if (!error && data && isMessageValid(data.concrete_notification)) {
+      setMessage(data.concrete_notification)
+    }
+  }, [data, error])
```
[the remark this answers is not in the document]
jeffersondarcy: useConcreteNotification uses loading now

src/hooks/useConcreteNotification.ts
jeffersondarcy: fixed

src/hooks/useConcreteNotification.ts
jeffersondarcy: isMessageValid was not checking for undefined and I had an error when the backend was down, but now it should be unnecessary

PR #31, opened 2022-11-29
Notification Preferences ControlPanel

LomyW:
Notification Preferences
System and Marketing Tabs


src/components/Notification/preferences/PreferenceItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+  Spacer,
+  Switch,
+  useBreakpointValue,
+  Tooltip
+} from 'native-base'
+import { useEffect, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import { getDataForNotificationPreference } from '../../../helper/notification-helper'
+
+type PrefProps = {
```
jeffersondarcy, line 17: in future we would also need data for the tool-tip or modal, I think it should be unproblematic to add later but we just need to consider this

src/components/Notification/preferences/PreferenceItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+  id: string
+  channel: { [channel: string]: boolean }
+}
+
+const PreferenceItem: React.FC<PrefProps> = ({ id, channel }) => {
+  const [emailActivated, setEmailActivated] = useState<boolean>(true)
+  const [channels, setChannels] = useState<string[]>([])
```
jeffersondarcy, line 24: you are never using channels, is this for a future mutation?

src/helper/notification-helper.tsx
```diff
@@ -51,6 +58,50 @@ function getIcon(messageType: string): ReactElement {
   }
 }
 
+type NotificationPreference = { title: string; icon: JSX.Element }
+
+const getDataForNotificationPreference = (
```
jeffersondarcy, line 63: this file is a bit confusing, would it be possible to have a reusable mapping structure, not inside a function for a specific use-case? Also mixing "notification" and "message" wording is confusing, but I am not sure if it can be avoided - I had a similar mixing

src/components/Notification/preferences/PreferenceItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const { t } = useTranslation()
+
+  const preference = getDataForNotificationPreference(id)
+
+  const isMobile = useBreakpointValue({
```
jeffersondarcy, line 29: can we do something reusable with useBreakpointValue so we don't need to duplicate this code in the codebase?

src/components/Notification/preferences/PreferencesData.tsx
```diff
@@ -0,0 +1,64 @@
+export const preferencesData = [
```
jeffersondarcy, line 1: this is currently not used, right?

src/components/Notification/preferences/PreferenceItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+    base: 340,
+    lg: '100%'
+  })
+
+  // TODO mutate preferences
+  const handleToggle = () => {
+    setEmailActivated(!emailActivated)
+  }
+
+  useEffect(() => {
```
jeffersondarcy, line 49: the logic is not easy to comprehend. can we may be move the logic out of the small component or at least create named methods where one can understand what logic it performs?

src/components/Notification/preferences/PreferenceItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { useEffect, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import { getDataForNotificationPreference } from '../../../helper/notification-helper'
+
+type PrefProps = {
+  id: string
+  channel: { [channel: string]: boolean }
+}
+
+const PreferenceItem: React.FC<PrefProps> = ({ id, channel }) => {
```
jeffersondarcy, line 22: what does `channel` represent? is this for the case we have tabs for different channels?

src/helper/notification-helper.tsx
```diff
@@ -51,6 +58,50 @@ function getIcon(messageType: string): ReactElement {
...
+  notificationPreference: string
```
jeffersondarcy, line 64: `notificationPreference` is not quite clear here - messageType or notificationType might be better. Preference for me would mean channel like "email" or "platform"

src/helper/notification-helper.tsx
```diff
@@ -94,4 +145,4 @@ const getTimeText = (timestamp: string): TimeText | string => {
   }
   return timeAsString
 }
-export { getIcon, getTimeText }
+export { getIconForMessageType, getDataForNotificationPreference, getTimeText }
```
jeffersondarcy, line 148: I don't like `getDataForNotificationPreference` something like `getDataForMessageTypeNotificationPreference` would be better understandable, but feel free to find a shorter name :)

src/pages/AppointmentsArchive.tsx
```diff
@@ -16,7 +16,7 @@ import {
 import { useNavigate } from 'react-router-dom'
 import { useTranslation } from 'react-i18next'
 import WithNavigation from '../components/WithNavigation'
-import NotificationAlert from '../components/Notification/NotificationAlert'
+import NotificationAlert from '../components/notification/NotificationAlert'
```
jeffersondarcy, line 19: I think capital letter Notification directory was a better choice, because the Component name also starts with capital. My IDE complains about something here - I think it's because of the casing

src/components/Notification/preferences/PreferenceItem.tsx
LomyW: yes, but naming is not ready now. I will change this.

LomyW: yes, but naming is not ready now. I will change this.

src/components/Notification/preferences/PreferenceItem.tsx
LomyW: already removed. :)

src/components/Notification/preferences/PreferenceItem.tsx
LomyW: I use it as it was used before.

Jonasdoubleyou:
The diff is completely wrong, looks like it reverts all changes on main @jeffersondarcy ?

LomyW (commented):
well done 👍

jeffersondarcy:
> The diff is completely wrong, looks like it reverts all changes on main @jeffersondarcy ?

thanks @Jonasdoubleyou, not sure when it happened - I will try to rebase

jeffersondarcy:
replaced with https://github.com/corona-school/user-app/pull/68

PR #36, opened 2022-12-04
Typed GraphQL Queries

Jonasdoubleyou:
To use
(1) Change the `gql` import to `import { gql } from "src/gql";`
(2) Replace tagged template literals with regular function calls ( `gql'...'` -> `gql('...')`)

Usage is optional at the moment as refactoring all queries is quite some work.


jeffersondarcy (approved):
looks okay for me, but I don't really have background knowledge

PR #42, opened 2022-12-08
Add Storybook

Jonasdoubleyou:
Adds an optional installation of Storybook, which developers can install on their machine.
Resolves https://github.com/corona-school/project-user/issues/517


package.json
```diff
@@ -82,15 +83,31 @@
...
+    "@storybook/addon-actions": "^6.5.14",
+    "@storybook/addon-essentials": "^6.5.14",
+    "@storybook/addon-interactions": "^6.5.14",
+    "@storybook/addon-links": "^6.5.14",
+    "@storybook/builder-webpack5": "^6.5.14",
+    "@storybook/manager-webpack5": "^6.5.14",
+    "@storybook/react": "^6.5.14",
+    "@storybook/testing-library": "^0.0.13"
+  },
   "scripts": {
```
timidak, line 96: One general point I have is that we could always generate a Storybook build in GitHub Actions and upload it to GitHub Pages. This would make it easier for the whole team (Designer, Marketing & Co.) to see the current stories without a development setup. What do you think?

Here's a snippet of the relevant scripts in the `package.json` file from anther project where we do that.
```
"predeploy": "build-storybook",
"deploy-storybook": "gh-pages -d storybook-static",
```

.storybook/preview.tsx
```diff
@@ -0,0 +1,22 @@
...
+  actions: { argTypesRegex: "^on[A-Z].*" },
+  controls: {
+    matchers: {
+      color: /(background|color)$/i,
+      date: /Date$/,
+    },
+  },
+}
+
+// TODO: Unfortunately the following does not work for MDX files?
```
timidak, line 15: You're right [according to the docs](https://storybook.js.org/docs/react/writing-docs/mdx#decorators-and-parameters) you should use the `<Meta />` tag to add decorators and it doesn't work globally afaics.

.storybook/main.js
```diff
@@ -0,0 +1,29 @@
...
+    "@storybook/addon-essentials",
+    "@storybook/addon-interactions"
+  ],
+  "framework": "@storybook/react",
+  "core": {
+    "builder": "@storybook/builder-webpack5"
+  },
+  "webpackFinal": (storybookConfig) => {
+    const appConfig = require("./../config/webpack.config.js")('development');
+    // This hacks in support for SCSS, Typescript and React Native module resolution into the Webpack configuration of Storybook
```
timidak, line 18: Nice!

timidak (approved):
Just some comments, feel free to refactor or merge. 😄  
It works on my machine without any problems. 👍🏼

package.json
```diff
@@ -82,15 +83,31 @@
...
     "update-backend": "graphql-codegen init",
-    "analyze-bundle": "npm run build --profile && npx source-map-explorer ./build/static/js/main.*.js --html ./build/source-map.html",
     "fetch-api": "graphql-codegen --config codegen.ts",
-    "check-types": "tsc"
+    
+    "analyze-bundle": "npm run build --profile && npx source-map-explorer ./build/static/js/main.*.js --html ./build/source-map.html",
+    
+    "dev-install": "npm install --include=optional",
+    "dev-storybook": "start-storybook -p 6006",
+    "dev-storybook-build": "build-storybook"
```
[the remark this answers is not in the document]
Jonasdoubleyou: Yeah that was intentional, meaning _you need `dev-install` before you can run all the `dev-` commands ... But yeah, not quite obvious

package.json
```diff
@@ -82,15 +83,31 @@
...
     "serve": "node serve.js",
+    
     "update-backend": "graphql-codegen init",
-    "analyze-bundle": "npm run build --profile && npx source-map-explorer ./build/static/js/main.*.js --html ./build/source-map.html",
     "fetch-api": "graphql-codegen --config codegen.ts",
-    "check-types": "tsc"
+    
+    "analyze-bundle": "npm run build --profile && npx source-map-explorer ./build/static/js/main.*.js --html ./build/source-map.html",
+    
+    "dev-install": "npm install --include=optional",
```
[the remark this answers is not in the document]
Jonasdoubleyou: good point

.storybook/preview.tsx
Jonasdoubleyou: Aaah, decorators only apply to stories, so inside MDX the components need to be wrapped in `<Story>...</Story>` ... Then the decorators are applied

package.json
Jonasdoubleyou: Very good idea, I'll try this in a followup PR

PR #59, opened 2022-12-18
Set error policy of login mutation to 'all' (project-user#526)

lsc0498:
@Jonasdoubleyou this solves the issue for now but I don't really understand why the login mutation threw an error in the first place. Shouldn't it just have been written into the error object of `loginResult` even with the default error policy?


Jonasdoubleyou:
Maybe its thrown as an internal server error with a non-200 status code?

Jonasdoubleyou (approved):

lsc0498:
Yes probably, I was just wondering why the effect of setting the errorPolicy is as it is 😅

PR #72, opened 2022-12-20
Removed edit feature from name component of student profile

realmayus:
See https://github.com/corona-school/project-user/issues/538

Also fixed naming of pupil profile to to ProfilePupil be more consistent with ProfilStudent


Jonasdoubleyou (approved):

Jonasdoubleyou:
If we consider changing the name for students a security problem, we might also want to prohibit this in the backen (for non elevated users, as we do with changing the mail), or @lsc0498 ?

realmayus:
Yeah, I already implemented that [here](https://github.com/corona-school/backend/pull/446) 🙂

PR #84, opened 2022-12-27
Promote course button

LomyW:
Button for helpers, to promote a course again!
[Works with this [Backend](https://github.com/corona-school/backend/tree/feat/50-promote-course) ]


src/pages/SingleCourse.tsx
```diff
@@ -978,6 +979,11 @@ const SingleCourse: React.FC = () => {
                             {t('single.global.clockFrom')} {Utility.formatDate(subcourse?.lectures[0].start)} {t('single.global.clock')}
                         </Text>
                     )}
+                    <Box my={2}>
+                        {subcourse && getTrafficStatus(subcourse!.participantsCount, subcourse!.maxParticipants) !== 'full' && (
```
jeffersondarcy, line 983: maybe move this check to a function "shouldShowPromoteButton" or "isSubcourseNotFull" - same code is used in the next box

src/widgets/PromoteButton.tsx
```diff
@@ -0,0 +1,86 @@
...
+    useEffect(() => {
+        if (!alreadyPromoted && capacity < 0.75 && isPublishedThreeDaysAgo(publishedAt)) {
+            setIsPromotionButtonDisabled(false);
+        } else {
+            setIsPromotionButtonDisabled(true);
+        }
+    }, [alreadyPromoted, capacity, publishedAt, daysDiff]);
+
+    return (
+        <div>
```
jeffersondarcy, line 74: shouldn't this div be inside the condition and not rendered empty?

timidak (changes requested):
Some changes to be made 🙂

src/pages/SingleCourse.tsx
```diff
@@ -866,9 +872,28 @@ const SingleCourse: React.FC = () => {
         { variables: { subcourseId } }
     );
 
+    const [promote, _promote] = useMutation(
```
timidak, line 875: You can just use destructuring here
```suggestion
    const [promote, { data, loading, error }] = useMutation(
```

src/pages/SingleCourse.tsx
```diff
@@ -866,9 +872,28 @@ const SingleCourse: React.FC = () => {
...
+        { variables: { subcourseId: subcourseId } }
+    );
+
     const { subcourse } = data ?? {};
     const { course } = subcourse ?? {};
 
+    const doPromote = async () => {
+        await promote();
+        if (_promote.error) {
+            toast.show({ description: 'Kurs konnte nicht beworben werden' });
```
timidak, line 890: This should be moved to the translated strings

src/widgets/PromoteButton.tsx
```diff
@@ -0,0 +1,57 @@
...
+    const isPublishedThreeDaysAgo = (publishDate: string): boolean => {
+        const { daysDiff } = getTimeDifference(publishDate);
+        if (publishDate === null || daysDiff > 3) {
+            return true;
+        }
+        return false;
+    };
+
+    useEffect(() => {
+        if (!alreadyPromoted && capacity < 0.75 && isPublishedThreeDaysAgo(publishedAt)) {
```
timidak, line 37: I would wrap this in a function and just call it on the isDisabled prop on the button. 
Imho is no need for an effect here.

timidak (changes requested):
Two minor changes 😄

src/widgets/PromoteButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+        }
+        return false;
+    };
+
+    const canPromoteCourse = () => {
+        if (!alreadyPromoted && capacity < 0.75 && isPublishedThreeDaysAgo(publishedAt)) {
+            return false;
+        } else {
+            return true;
+        }
```
timidak, line 37: hence the value in the if is a boolean you can just return it 😄 
To achieve the same behavior, you'll have to negate it.
```suggestion
        return !(!alreadyPromoted && capacity < 0.75 && isPublishedThreeDaysAgo(publishedAt)) 
```

timidak (approved):
LGTM 🎉

src/Utility.ts
```diff
@@ -128,7 +128,8 @@ export const getFirstLectureFromSubcourse: (lectures: LFLecture[], pastLectures?
 };
 
 export const getTrafficStatus: (participants: number, maxParticipants: number) => TrafficStatus = (participants = 0, maxParticipants = 0) => {
-    return participants === maxParticipants ? 'full' : maxParticipants - participants <= 5 ? 'last' : 'free';
+    const status = participants >= maxParticipants ? 'full' : maxParticipants - participants <= 5 ? 'last' : 'free';
```
jeffersondarcy, line 131: could be done without variable like the previous version

src/pages/SingleCourse.tsx
```diff
@@ -978,6 +1003,19 @@ const SingleCourse: React.FC = () => {
                             {t('single.global.clockFrom')} {Utility.formatDate(subcourse?.lectures[0].start)} {t('single.global.clock')}
                         </Text>
                     )}
+                    <Box my={2}>
+                        {subcourse && (
+                            <PromoteButton
```
jeffersondarcy, line 1008: I think it would be better to check the conditions before rendering the button and just tell the button if it should be enabled or disabled

src/widgets/PromoteButton.tsx
```diff
@@ -0,0 +1,29 @@
+import { Button, Tooltip, useBreakpointValue, useTheme } from 'native-base';
+import { useTranslation } from 'react-i18next';
+
+type Props = {
+    isDisabled: boolean | undefined;
+    loading: boolean;
+    promote: () => void;
+};
+
+const PromoteButton: React.FC<Props> = ({ isDisabled, loading, promote }) => {
```
jeffersondarcy, line 10: do you need to pass loading, or can you apply isDisabled logic before

jeffersondarcy: I would rename promote to something generic like onClick

src/pages/SingleCourse.tsx
```diff
@@ -866,9 +873,41 @@ const SingleCourse: React.FC = () => {
...
+
+    const isPublishedThreeDaysAgo = (publishDate: string): boolean => {
+        const { daysDiff } = getTimeDifference(publishDate);
+        if (publishDate === null || daysDiff > 3) {
+            return true;
+        }
+        return false;
+    };
+
+    const cannotPromoteCourse = () => {
```
jeffersondarcy, line 906: why don't you check subcourse, subcourse.published and loading here? than you would only have one piece of logic

LomyW: thought of not displaying the button if the subcourse isn't published - not only disable the button.

src/pages/SingleCourse.tsx
```diff
@@ -866,9 +873,41 @@ const SingleCourse: React.FC = () => {
...
+        if (!subcourse) return;
```
jeffersondarcy, line 907: maybe !subcourse || subcourse.published should still return false, just in case it is not checked in render?

jeffersondarcy (approved):

jeffersondarcy (approved):

PR #85, opened 2022-12-27
Feat/LF-76 activate/deactivate all 

jeffersondarcy:
Buttons to activate and deactivate all switches in Marketing Notifications

### DIFF
--- a/src/components/notifications/preferences/ToggleAll.tsx
+++ b/src/components/notifications/preferences/ToggleAll.tsx
@@ -1,0 +1,60 @@
+import { Box, HStack, VStack, Text, Spacer, Switch, useBreakpointValue } from 'native-base';
+import { useTranslation } from 'react-i18next';
+import { NotificationCategories } from '../../../helper/notification-preferences';
+import { useContext, useEffect } from 'react';
+import { NotificationPreferencesContext } from '../../../pages/notification/NotficationControlPanel';
+
+type PrefProps = {
+    notificationCategories: NotificationCategories;
+};
+
+export const ToggleAll: React.FC<PrefProps> = ({ notificationCategories }) => {
+    const { userPreferences, updateUserPreference, channels } = useContext(NotificationPreferencesContext);
+    const { t } = useTranslation();
+
+    const maxW = useBreakpointValue({
+        base: 200,
+        lg: '100%',
+    });
+
+    const width = useBreakpointValue({
+        base: 340,
+        lg: '100%',
+    });
+
+    const isAllDisabled = () => {
+        Object.keys(notificationCategories).forEach((category: string) =>
+            channels.every((channel: string) => console.log(category, channel, userPreferences[category][channel]))
+        );
+        return true;
+    };
+    /*
+    const isAllDisabled = () =>
+        Object.keys(notificationCategories).every((category: string) => channels.every((channel: string) => userPreferences[category][channel] === false));
+*/
+    useEffect(() => {
+        console.log(userPreferences);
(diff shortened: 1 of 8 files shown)


src/components/notifications/preferences/ToggleAll.tsx
```diff
@@ -0,0 +1,76 @@
...
+        base: 340,
+        lg: '100%',
+    });
+
+    // use isEnabled=true to check if all preferences are enabled and isEnabled=false to check if all preferences are disabled
+    const isAllEnabledOrDisabled = (isEnabled: boolean = true) => {
+        return Object.keys(notificationCategories).every((category: string) =>
+            channels.every((channel: string) => {
+                if (!userPreferences.hasOwnProperty(category) || !userPreferences[category].hasOwnProperty(channel)) return true;
+                console.log(category, channel, userPreferences[category][channel]);
```
Jonasdoubleyou, line 33: Debug logging? Either remove or add some context to the logline

src/components/notifications/preferences/ToggleAll.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { NotificationPreferencesContext } from '../../../pages/notification/NotficationControlPanel';
+import { getAllPreferencesInCategorySetToValue } from '../../../helper/notification-helper';
+
+type PrefProps = {
+    notificationCategories: NotificationCategories;
+};
+
+export const ToggleAll: React.FC<PrefProps> = ({ notificationCategories }) => {
+    const { userPreferences, channels, updateUserPreferences } = useContext(NotificationPreferencesContext);
+    const allEnabled = useRef(false);
```
Jonasdoubleyou, line 14: Why is this a ref? 

const allEnabled = useMemo(isAllEnables, [...]);

jeffersondarcy: this value is recalculated in useEffect every time any switch is changed and is set true when all switches are enabled.  It's used as state, but without rerendering. Do you think useMemo is better suited for this?

Jonasdoubleyou: > It's used as state, but without rerendering
Yes, but you actually need to rerender as `allEnabled.current` is used as a prop. This only works as the component is rerendered anyways whenever `allEnabled` is changed. That's quite dangerous.

Actually, `allEnabled` is not a state itself, but can be derived from another state - `userPreferences` - that's exactly what useMemo is good for - deriving complex calculations from state.

jeffersondarcy: okay, I will try to incorporate this - I don't have experience with useMemo. can you explain the danger - the value is recalculated in useEffect always prior to setting state&rerender

src/components/notifications/preferences/ToggleAll.tsx
jeffersondarcy: removed

src/components/notifications/preferences/ToggleAll.tsx
Jonasdoubleyou: Well you can write to allEnabled.current, and the UI might not rerender. So the "state" can diverge from what the user sees.

jeffersondarcy: done

jeffersondarcy: sorry, forgot to push. now is updated

Jonasdoubleyou (approved):

PR #88, opened 2022-12-29
add missing keys

jeffersondarcy:
I have added keys for components inside map to avoid warnings

### DIFF
--- a/src/modals/DeactivateAccountModal.tsx
+++ b/src/modals/DeactivateAccountModal.tsx
@@ -104,7 +104,9 @@ const DeactivateAccountModal: React.FC<Props> = ({ isOpen, onCloseModal }) => {
                         <Radio.Group name="reasons" onChange={setReason}>
                             <VStack space={space['0.5']}>
                                 {reasons.map((_, index: number) => (
-                                    <Radio value={`${index + 1}`}>{t(`profile.Deactivate.${userType}.${index + 1}`)}</Radio>
+                                    <Radio key={index} value={`${index + 1}`}>
+                                        {t(`profile.Deactivate.${userType}.${index + 1}`)}
+                                    </Radio>
                                 ))}
                             </VStack>
                         </Radio.Group>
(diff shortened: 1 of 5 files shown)


src/modals/DeactivateAccountModal.tsx
```diff
@@ -104,7 +104,9 @@ const DeactivateAccountModal: React.FC<Props> = ({ isOpen, onCloseModal }) => {
                         <Radio.Group name="reasons" onChange={setReason}>
                             <VStack space={space['0.5']}>
                                 {reasons.map((_, index: number) => (
-                                    <Radio value={`${index + 1}`}>{t(`profile.Deactivate.${userType}.${index + 1}`)}</Radio>
+                                    <Radio key={index} value={`${index + 1}`}>
```
jeffersondarcy, line 107: had to use index here, feel free to suggest something better

src/modals/DissolveMatchModal.tsx
```diff
@@ -34,7 +34,9 @@ const DissolveMatchModal: React.FC<DissolveModalProps> = ({ showDissolveModal, o
                             <Radio.Group name="dissolve-reason" value={reason} onChange={setReason}>
                                 <VStack space={space['1']}>
                                     {reasons.map((_: number, index: number) => (
-                                        <Radio value={`${index + 1}`}>{t(`matching.dissolveReasons.${userType}.${index + 1}`)}</Radio>
+                                        <Radio key={index} value={`${index + 1}`}>
```
jeffersondarcy, line 37: didn't find anything better than index here

src/pages/pupil/Dashboard.tsx
```diff
@@ -319,7 +319,7 @@ const Dashboard: React.FC<Props> = () => {
                                         if (!course) return <></>;
 
                                         return (
-                                            <Column minWidth="230px" maxWidth="300px" flex={1} h="100%">
+                                            <Column minWidth="230px" maxWidth="300px" flex={1} h="100%" key={`${course.course.description}+${lecture.start}`}>
```
jeffersondarcy, line 322: id was not working here @Jonasdoubleyou I don't think the line `navigate(`/single-course/${course.id}`);` below works

Jonasdoubleyou: Well, there can be multiple appointments (aka lectures) for the same subcourse. The page is for a subcourse though, so multiple appointments will open the same subcourse page.

Jonasdoubleyou (approved):

src/pages/pupil/Dashboard.tsx
jeffersondarcy: I have tried to use course.id for index, but it was apparently undefined. Maybe it works in some other branch

PR #89, opened 2022-12-29
LF-33 notification link

jeffersondarcy:
Notification is now clickable and navigation resolves internal and external routes

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -12,22 +12,25 @@ import { LernfairProvider } from './hooks/useLernfair';
 import { IssueReporter } from './IssueReporter';
 import { NotificationsProvider } from './components/NotificationsProvider';
 import { ToastNotifications } from './components/ToastNotifications';
+import { BrowserRouter } from 'react-router-dom';
 
 function App() {
     return (
         <LernfairProvider>
             <LFModalProvider>
                 <LFApolloProvider>
-                    <NativeBaseProvider theme={Theme}>
-                        <IssueReporter>
-                            <MatomoProvider value={matomo}>
-                                <NotificationsProvider>
-                                    <Navigator />
-                                    <ToastNotifications />
-                                </NotificationsProvider>
-                            </MatomoProvider>
-                        </IssueReporter>
-                    </NativeBaseProvider>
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -12,22 +12,25 @@ import { LernfairProvider } from './hooks/useLernfair';
 import { IssueReporter } from './IssueReporter';
 import { NotificationsProvider } from './components/NotificationsProvider';
 import { ToastNotifications } from './components/ToastNotifications';
+import { BrowserRouter } from 'react-router-dom';
```
timidak, line 15: This will clash with React Native. Iirc @Jonasdoubleyou mentioned that in the first meetings

timidak: Nevermind, I just saw that the BrowserRouter was already implemented 😄

Jonasdoubleyou: Yeah, c.f. https://github.com/corona-school/project-user/issues/552 ... Maybe we'll never build a full native app with react native, just a "more native PWA" ... Still, I wanted to keep the option open to somewhen go native without too much effort, so the usage of BrowserRouter here is fine ...

Jonasdoubleyou (approved):

src/components/notifications/MessageBox.tsx
```diff
@@ -1,65 +1,91 @@
...
+    const navigateToLink = () => {
+        if (typeof navigateTo !== 'string') return null;
+        if (navigateTo.charAt(0) === '/') {
+            return navigate(navigateTo);
+        }
+        return navigateExternal();
+    };
 
-export default MessageBox
+    const navigateExternal = () => window.open(navigateTo, '_blank');
```
timidak, line 42: I'm not sure how a PWA would react on this event. But it should work for now.

timidak (approved):
LGTM 👍🏼

src/components/notifications/MessageBox.tsx
Jonasdoubleyou: @timidak we can worry about the PWA / native topics once the User App is live and has notifications, appointments, ... ;)

PR #95, opened 2023-01-16
Refactor toast position

LomyW:


timidak (approved):
The history seems to be off, but idk why. 
Also, it might have been easier to use, if we had provided a custom hook `useToast()` where placement top is set as standard. 
But for now it does the job. 👍🏼

PR #96, opened 2023-01-17
LF-60 update query

jeffersondarcy:
minor change because of field renaming in https://github.com/corona-school/backend/pull/455


LomyW (approved):
👍

PR #101, opened 2023-01-19
Add: active matches first on pupils and students

Leo310:
Resolves: corona-school/project-user#564


Leo310:
Not sure what that GraphQL verification error should be

Jonasdoubleyou (approved):
Can be ignored. That's a change of mine that hasn't reached production yet. The check is just a reminder to deploy the backend before the frontend as there is a dependency.

PR #107, opened 2023-01-25
handle incomplete notification messages

jeffersondarcy:


Jonasdoubleyou:
hm, how is it an error state when the message is null? That can happen i.e. for the Notification 'Login via Email'

LomyW (approved):
Looks good! 🎉

jeffersondarcy:
> hm, how is it an error state when the message is null? That can happen i.e. for the Notification 'Login via Email' ... Or in case you want to filter out these somewhere else, I guess it's fine to just fail here by throwing an error, this state should then not be reached here

I am not sure what exactly you mean. Should the message be displayed or not, What should I change?

Jonasdoubleyou (approved):
Yeah this way it should be okay.

Jonasdoubleyou (approved):

PR #108, opened 2023-01-26
fix: websocket reconnect

timidak:
- added reconnect interval
- improved logging


src/lib/Websocket.ts
```diff
@@ -1,38 +1,63 @@
...
+            this.timerId = undefined;
+
+            this.socket!.onclose = (event) => {
+                log('WebSocket', 'Connection closed...');
+                debug('WebSocket', `Code: ${event.code} - reason: ${event.reason}`);
+                this.isConnected = false;
+                this.socket = null;
+
+                // try to reconnect after connection was closed abnormally (cf. https://www.rfc-editor.org/rfc/rfc6455.html#section-7.4.1)
+                if (event.code === 1006 && !this.timerId) {
```
jeffersondarcy, line 45: are we sure, we only need to handle code 1006? is there any other possible disconnect reason we need to handle with re-connect?

src/lib/Websocket.ts
```diff
@@ -1,38 +1,63 @@
...
+                    this.timerId = window.setInterval(() => {
+                        this.connect(url);
```
jeffersondarcy, line 47: does it work? connect returns null if this.socket !== null

src/lib/Websocket.ts
timidak: For now, I would say this should be the one and case, where the front-end should try to reconnect. The other cases are normally handled by the protocol and I wouldn't reconnect, when a server side error happened. @jeffersondarcy

PR #113, opened 2023-01-26
Fix key and type errors

jeffersondarcy:


timidak:
Closed after quick check with @jeffersondarcy. It took longer than expected and should be started freshly in the future if needed.

PR #115, opened 2023-01-27
Fix: tuned condition of welcome message appearance

Leo310:
Resolves: corona-school/project-user#520

### DIFF
--- a/src/widgets/ImportantInformation.tsx
+++ b/src/widgets/ImportantInformation.tsx
@@ -57,6 +57,9 @@ const ImportantInformation: React.FC<Props> = ({ variant }) => {
 			  subjectsFormatted {
 				name
 			  }
+              subcoursesJoined {
+                  id
+              }
 			  matches {
 				dissolved
 				createdAt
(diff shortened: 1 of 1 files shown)


src/widgets/ImportantInformation.tsx
```diff
@@ -115,7 +118,8 @@ const ImportantInformation: React.FC<Props> = ({ variant }) => {
         (student?.canCreateCourse?.reason === 'not-instructor' && student.canRequestMatch?.reason === 'not-tutor')
     )
         infos.push({ label: 'kennenlernen', btnfn: [() => window.open(process.env.REACT_APP_SCREENING_URL)], lang: {} });
-    if (pupil && !pupil?.firstMatchRequest) infos.push({ label: 'willkommen', btnfn: [() => navigate('/group'), () => navigate('/matching')], lang: {} });
+    if (pupil?.firstMatchRequest !== null && !pupil?.firstMatchRequest && pupil?.subcoursesJoined && pupil?.subcoursesJoined.length == 0)
```
Jonasdoubleyou, line 121: How can firstMatchRequest be not null but false? Also this won't be set for accounts which requested a match before the timestamp was introduced, so better also check matches

Leo310: not sure but on the test students, firstMatchRequest also wasn't set. So I thought I would be on the safe side if I just check for null

Jonasdoubleyou: Yeah but then !firstMatchRequest should be enough no?

Leo310: oh true

Jonasdoubleyou (approved):

PR #119, opened 2023-01-30
Appointments

LomyW:


public/index.html
```diff
@@ -1,35 +1,38 @@
...
+        <link rel="icon" type="image/png" sizes="16x16" href="/favicons/favicon-16x16.png" />
+        <link rel="manifest" href="/site.webmanifest" />
+        <link rel="mask-icon" href="/favicons/safari-pinned-tab.svg" color="#5bbad5" />
+        <link rel="shortcut icon" href="/favicons/favicon.ico" />
+        <meta name="msapplication-TileColor" content="#da532c" />
+        <meta name="msapplication-config" content="/favicons/browserconfig.xml" />
+        <meta name="theme-color" content="#ffffff" />
+
+        <link type="text/css" rel="stylesheet" href="https://source.zoom.us/2.11.5/css/bootstrap.css" />
+        <link type="text/css" rel="stylesheet" href="https://source.zoom.us/2.11.5/css/react-select.css" />
```
Jonasdoubleyou, line 21: Can we use `rel="preload"` here ([doc](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload)? It feels wrong to use our initial loading time for stuff that is used way later ...

src/types/lernfair/User.ts
```diff
@@ -1,5 +1,15 @@
+import { AppointmentParticipant, Organizer } from '../../gql/graphql';
 import { State } from './State';
 
+export enum UserType {
+    STUDENT = 'student',
+    PUPIL = 'pupil',
+}
+
+export enum AttendanceStatus {
```
Jonasdoubleyou, line 9: (maybe a better fit for Appointment.ts?)

src/pages/Chat.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
```
Jonasdoubleyou, line 1: Isn't this added in the feat/integrated-chat branch?

Jonasdoubleyou (approved):

PR #122, opened 2023-02-04
Desktop-View: Match Process

Leo310:
Start of resolving: [#530](https://github.com/corona-school/project-user/issues/530)

Implemented the same previous and next buttons as on the registration page.
Atm only for the student match process.

**Open Question:**
@tobiasbork (@Jonasdoubleyou ) How should this look like in desktop view? Maybe have more than 2 subjects in one row?
image


Jonasdoubleyou:
yes, sounds good

Leo310:
@tobiasbork @Jonasdoubleyou implemented it, but I am not sure if this is the best most responsive solution. But it seems to be working. Then I would also implement it for the pupil match process.

Jonasdoubleyou:
I like it! With https://github.com/corona-school/user-app/pull/123 I introduced a component <SubjectSelector> that is reused across all flows. Maybe you could cherry-pick your changes onto that change?

Leo310:
Will do!

Jonasdoubleyou (approved):
nice

PR #158, opened 2023-02-25
Add: working prototype of configurable-important-informations

Leo310:
Resolves other part of corona-school/project-user#609.


Leo310:
Ah fck, something weird happened.

PR #166, opened 2023-02-28
Refactor alerts based on backend reasons.

Maascha:


src/lang/lernfair/de.ts
```diff
@@ -84,37 +84,36 @@ const lernfair = {
     schoolclass: '{{class}}. Klasse',
 
     reason: {
-        'subcourse-full': {
-            coursetext: 'Dieser Kurs ist leider grade ausgebucht.',
+        course: {
+            // from backend/graphql/student/fields.ts
```
Maascha, line 88: Do we have something like "if this then that" to make sure the text stays in sync to the backend?

src/lang/lernfair/de.ts
```diff
@@ -84,37 +84,36 @@ const lernfair = {
...
+            instructor: {
+                'not-instructor': 'Du bist noch nicht als Kursleiter:in registriert.',
+                'not-screened': 'Du bist noch nicht als Kursleter:in freigeschaltet. Bitte führe zuerst dein Kennenlerngespräch mit uns.',
```
Jonasdoubleyou, line 91: typo

src/lang/lernfair/de.ts
Jonasdoubleyou: Nope, most semantic changes in the Backend will cause changes to the GraphQL Schema, which is part of the build process of the user app. The reason is a string though unfortunately. We could rework the Backend to use different subclasses of Reason, each with it's own enum for the reason field, then we could use mapped typed in Typescript here ... But that would need quite some rework ...

src/lang/lernfair/de.ts
Maascha: done

Jonasdoubleyou (approved):

PR #168, opened 2023-03-01
Add Lern-Fair Live test run information for student and pupils.

Maascha:
1. ImportantInformation cards can have <b> and <br> tags.
2. Text modified for the test run. Pupils with open Match requests and screened tutors should see these cards.

Should only be PR after the /hausaufgabenhilfe-anmeldung forwards to the correct zoom registration link.

### DIFF
--- a/src/lang/helperwizard/de.ts
+++ b/src/lang/helperwizard/de.ts
@@ -29,7 +29,7 @@ const helperwizard = {
     interestconfirmation: {
         title: 'Brauchst du noch Unterstützung?',
         content:
-            'Du bist nun oben auf unserer Warteliste und wir können dich bald mit einem:r Lernpartner:in verbinden. Brauchst du immer noch Hilfe in den Fächern {{subjectSchüler}}. Wenn du “Nein” klickst oder dich nicht innerhalb von 14 Tagen bei uns meldest, vergeben wir deinen Platz an andere Schüler:innen.',
+            'Du bist nun oben auf unserer Warteliste und wir können dich bald mit einem:r Lernpartner:in verbinden. Brauchst du immer noch Hilfe in den Fächern {{subjectSchüler}}. Wenn du “Nein” klickst oder dich nicht innerhalb von  <b>14 Tagen </b> bei uns meldest, vergeben wir deinen Platz an andere Schüler:innen.',
         buttons: ['Ja', 'Nein'],
     },
     kontaktSchüler: {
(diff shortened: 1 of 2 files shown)


src/widgets/ImportantInformation.tsx
```diff
@@ -273,6 +273,21 @@ const ImportantInformation: React.FC<Props> = ({ variant }) => {
             });
         }
 
+        // -------- TEMP: Test of Homework help -----
+        if (roles.includes('TUTEE') && (pupil?.openMatchRequestCount ?? 0) > 0)
```
Jonasdoubleyou, line 277: Why only with open match request count?

src/lang/helperwizard/de.ts
```diff
@@ -76,5 +76,18 @@ const helperwizard = {
             'Es ist soweit, wir haben eine:n Schüler:in für dich gefunden! {{nameSchüler}} freut sich schon sehr auf deine Unterstützung. Bitte nehme Kontakt mit {{nameSchüler}} auf und vereinbare ein erstes Treffen.',
         buttons: ['Kontakt aufnehmen', 'Lernpartner:in ansehen'],
     },
+    // temporary
```
Jonasdoubleyou, line 79: What will a future reader do with this comment? I think this could be removed, or if a comment is needed a bit more context would be good (i.e. "test phase during march")

src/widgets/ImportantInformation.tsx
Maascha: Only tutees on the waiting list should receive this notification, to start with. We discussed to maybe open it to all tutees, in case there is not enough demand by waiting list pupils themselves. I would open a PR if we see this necessary around mid march

src/lang/helperwizard/de.ts
Maascha: well, best case this temporary comment will be only there during march, but I see your point and will change it in the next commit.

src/widgets/ImportantInformation.tsx
Jonasdoubleyou: Okay, otherwise this could just be created as an important notification via Retool

Jonasdoubleyou (approved):
Please wait with the merge till the link on the website works ...

src/widgets/ImportantInformation.tsx
Maascha: I thought important notifications via Retool are pop-ups, am I mistaken and they appear as "Nächste Schritte" cards

Jonasdoubleyou: Yes, see 283ff. ...

PR #172, opened 2023-03-02
Add legal to welcome and login page.

Maascha:


Jonasdoubleyou (approved):

Jonasdoubleyou (changes requested):
Hm, that breaks the layout of the welcome screen image

Jonasdoubleyou:
I think the problem is the 100vw combined with the padding that makes the box overflow. Maybe by using display='flex' on the container and flexGrow='1' on the footer the footer would always stay on the bottom without overflowing

Maascha:
Thanks, good catch. I added a flex container to welcome, but it think for login it worked before, or would you add `display="flex" `  there too?

Jonasdoubleyou (approved):

Jonasdoubleyou:
Yes, Login looks good

PR #182, opened 2023-03-08
Add: navigation to reset password page + minor change to reset logic

Leo310:
Resolves corona-school/project-user#567


Jonasdoubleyou:
What happens when users navigate to /reset-password directly (when they request a password reset on login)? Do we then also show a back button?

Leo310:
@Jonasdoubleyou Ah, I didn't think about that. Maybe we should provide different layouts? The old one (no back-button)  when users navigate directly to /reset-password and this one (with back-button) when they come from /settings. 
And then provide them on two different routes or pass another parameter to differentiate between them?

Jonasdoubleyou:
Two different routes sounds good, I guess sooner or later we want to have different wording for the two different flows anyways

Jonasdoubleyou (approved):

PR #185, opened 2023-03-10
Add dynamic description textfield

LomyW:


src/pages/course-creation/TextInput.tsx
```diff
@@ -0,0 +1,30 @@
+import { useState } from 'react';
+import { Input, useBreakpointValue } from 'native-base';
+
+interface InputProps {
+    value: string;
+    placeholder: string;
+    onChangeText: (text: string) => void;
+}
+
+const TextInput: React.FC<InputProps> = ({ value, placeholder, onChangeText }) => {
```
timidak, line 10: TextArea might be a better name I guess since it is a multiline text input?

src/pages/course-creation/TextInput.tsx
```diff
@@ -0,0 +1,30 @@
...
+        base: true,
+        lg: false,
+    });
+
+    return (
+        <Input
+            placeholder={placeholder}
+            value={value}
+            onChangeText={onChangeText}
+            style={{ height: Math.max(isMobile ? 100 : 300, height) }}
```
timidak, line 22: I'm not sure if this would work on React Native as well. But I'd keep it this way since we don't know if we ever switch to native.

timidak (commented):
Some suggestions from my side 😄

timidak (approved):
LGTM 🎉

PR #189, opened 2023-03-11
Adapt Matomo Config

Jonasdoubleyou:


Jonasdoubleyou:
Could one of you take over this PR? I have no idea of Matomo (and no account?) so I cannot verify whether the config works

Leo310 (approved):
I will take a look at it

PR #193, opened 2023-03-14
Waiting list

LomyW:


src/pages/single-course/Waitinglist.tsx
```diff
@@ -0,0 +1,116 @@
...
+            await joinFromWaitinglist({ variables: { subcourseId: subcourseId, pupilId: pupilId } });
+            setIsJoinPupilModalOpen(false);
+            toast.show({ description: t('single.waitinglist.toast'), placement: 'top' });
+            refetch();
+        },
+        [joinFromWaitinglist, refetch, subcourseId, toast]
+    );
+
+    const handleIncrease = useCallback(
+        async (newMaxParticiants: number) => {
```
timidak, line 67: The term `newMaxParticipants` seems to be semantically wrong. `new` would mean the new value not the value to add to the present value. `amountToBeAdded` or `participantsAmountToBeAdded` or sth. like that would be a better fit imho.

src/widgets/WaitinglistBanner.tsx
timidak, line 1: Where do you use this widget? I can't find the component where it is rendered out. 😄

src/assets/icons/lernfair/Icon_SandClock_small.svg
timidak, line 1: Not sure, but do we really need both svg's? Couldn't we just scale the svg when using it?

timidak (changes requested):
Some remarks from my side

src/assets/icons/lernfair/Icon_SandClock_small.svg
LomyW: not this easy, just took over the icons from the design, just did it the way it was done before.

src/widgets/WaitinglistBanner.tsx
LomyW: It is rendered in the `singleCourse` component, if a pupil is on the waitinglist. It replaces the `Kurs verlassen` Button

src/pages/SingleCourse.tsx
```diff
@@ -687,82 +681,34 @@ function PupilLeaveCourseAction({ subcourse, refresh }: { subcourse: Pick<Subcou
...
-                        leaveWaitingList();
-                    }}
-                    width={ButtonContainer}
-                    marginBottom={space['0.5']}
-                    isDisabled={loading}
-                >
-                    Warteliste verlassen
-                </Button>
+            <VStack space={space['0.5']} mb="5">
+                <WaitinglistBanner courseStatus={courseStatus} onWaitinglist={onWaitinglist} subcourseId={subcourse.id} refresh={() => refresh()} />
```
LomyW, line 700: @timidak here the banner is used 😄

src/types/lernfair/Course.ts
```diff
@@ -48,4 +49,14 @@ export type LFInstructor = {
...
 
+export type LFPupilOnWaitinglist = PupilOnWaitinglist[] | undefined;
+
+export type PupilOnWaitinglist = {
+    id: number;
+    firstname: string;
+    lastname: string;
+    schooltype: Pupil_Schooltype_Enum;
+    grade: string;
+};
```
timidak, line 60: Using the `Pick` utility type might be a better option here.

timidak (commented):
Just suggestions 👍🏻 
The rest of the code looks fine so far!

PR #195, opened 2023-03-15
Refactored session handling

timidak:
- Removed session handling from components
- Moved it to `useApollo` hook
- Added Alert in case of invalid token

Fixes
- https://github.com/corona-school/project-user/issues/578
- https://github.com/corona-school/project-user/issues/584

### DIFF
--- a/src/pages/Login.tsx
+++ b/src/pages/Login.tsx
@@ -3,7 +3,7 @@ import { gql } from '../gql';
 import { FetchResult, useMutation } from '@apollo/client';
 import Logo from '../assets/icons/lernfair/lf-logo.svg';
 
-import { Box, Button, Heading, Image, Modal, Row, Text, useBreakpointValue, useTheme, VStack, Link } from 'native-base';
+import { Box, Button, Heading, Image, Modal, Row, Text, useBreakpointValue, useTheme, VStack, Link, Alert, HStack, useToast, CloseIcon } from 'native-base';
 import useApollo from '../hooks/useApollo';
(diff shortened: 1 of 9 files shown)


src/pages/Login.tsx
```diff
@@ -68,6 +60,29 @@ export default function Login() {
 
     useEffect(() => {
         if (sessionState === 'logged-in') navigate(retainPath);
```
Jonasdoubleyou, line 62: When I navigate directly to /login while being logged in, retainPath is undefined and navigate(...) fails:
image

I guess we should assume that retainPath is optional and add a `?? '/start'` here

Jonasdoubleyou: Looks good apart from this

Jonasdoubleyou: Uh, we also make the same assumption that the state is set in other places, i.e. https://github.com/corona-school/user-app/blob/8b8bba5f9bc75cd25a935bcd3cb8f8818691c59d/src/pages/Welcome.tsx#L18 and this already affected 20 users:
image

Jonasdoubleyou: Can you directly fix this in all places tomorrow?

timidak: Sure, I'll fix them tomorrow in this PR, @Jonasdoubleyou. 🙂

LomyW (approved):
Looks good 🎉 and works smoothly!

Jonasdoubleyou:
I merged this to get the fix in ASAP ...

PR #196, opened 2023-03-16
Handle invalid mails during login

timidak:
- added package [validator.js](https://github.com/validatorjs/validator.js) instead of `class-validator` because `class-validator` uses `validator.js` under the hood
- added validation to `Login.tsx` and used the FormErrors
- refactored `PersonalData.tsx` to use the same validation as `Login.tsx`
- added translation text

Fixes:
- https://github.com/corona-school/project-user/issues/645

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
         "terser-webpack-plugin": "^5.2.5",
         "typescript": "^4.7.4",
         "ua-parser-js": "^1.0.32",
+        "validator": "^13.9.0",
         "web-vitals": "^2.1.4",
         "webpack": "^5.64.4",
         "webpack-dev-server": "^4.6.0",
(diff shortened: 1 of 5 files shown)


LomyW (approved):
Nice! 🥳

package.json
```diff
@@ -195,7 +196,8 @@
     },
     "dependencies": {
         "compression": "^1.7.4",
-        "express": "^4.18.2"
+        "express": "^4.18.2",
+        "validator": "^13.9.0"
```
Jonasdoubleyou, line 200: Isn't it a devDependency (not needed in the server at runtime)?

Jonasdoubleyou: yes it should be a devDependency.

Jonasdoubleyou (approved):

PR #203, opened 2023-03-21
Add validation for hour fields in cert request form

realmayus:
Closes https://github.com/corona-school/project-user/issues/651


Jonasdoubleyou:
Could you check whether we also support floating point here easily? Apparently people want to enter them, and the DB type is actually a Decimal (and in GraphQL its a Float)

Jonasdoubleyou (approved):

PR #209, opened 2023-03-25
Add: basic fileupload functionalities

Leo310:
Resolves part of corona-school/project-user#611

**Next steps:**
- Overall better error handling (e.g. backend returns error response)
- Better UI of attached files

**Open questions:**
- "Betreff" and "Deine Nachricht" should be mandatory and "Dateien Anhängen" optional correct?
- What file types should be allowed?
- Atm getting authorization error:
image
- Optional: Maybe support uploading multiple files at once in backend? Would be faster I think


Jonasdoubleyou:
> "Betreff" and "Deine Nachricht" should be mandatory and "Dateien Anhängen" optional correct?

correct.

> What file types should be allowed?

I don't think we should restrict that.

> Atm getting authorization error

With which testuser and which course? Pupils should only see the contact option when they are actually allowed to contact.

Leo310:
@Jonasdoubleyou it was for testuser pupil1 on course "Lorem Ipsum" :)

Jonasdoubleyou:
True that is a bug, fixed with https://github.com/corona-school/user-app/pull/210

Leo310:
@Jonasdoubleyou what do you think?
I think this is sufficient for now.

Jonasdoubleyou (commented):
Very nice! However for some reason it does not work when I contact participants of a course as an instructor. Maybe the files get lost somewhere? (Could be that this is a dev only problem, but we should check)

Jonasdoubleyou:
image

And maybe we can use flexDirection=column for smaller devices?

Jonasdoubleyou (approved):
nice!

PR #213, opened 2023-03-27
Adjust course card infos

LomyW:
https://github.com/corona-school/project-user/issues/595#issuecomment-1484141424


src/Utility.ts
```diff
@@ -133,6 +133,7 @@ export const getTrafficStatusText = (subcourse: { course: { courseState?: Course
     if (!subcourse.published && subcourse.course.courseState === Course_Coursestate_Enum.Created) return i18next.t('single.global.courseState.draft');
     if (!subcourse.published && subcourse.course.courseState === Course_Coursestate_Enum.Submitted) return i18next.t('single.global.courseState.submitted');
     if (!subcourse.published && subcourse.course.courseState === Course_Coursestate_Enum.Allowed) return i18next.t('single.global.courseState.draft');
+    if (subcourse.course.courseState === Course_Coursestate_Enum.Denied) return i18next.t('single.global.courseState.denied');
```
Jonasdoubleyou, line 136: All courseStates could be moved into if(!subcourse.published)

src/pages/subcourse/AllSubcourses.tsx
```diff
@@ -44,7 +44,7 @@ const AllSubcourses: React.FC<GroupProps> = ({ languageCourses, courses, focusCo
                 showStatus={(subcourse as any)?.isInstructor ?? false}
                 trafficLightStatus={getTrafficStatus(subcourse.participantsCount || 0, subcourse.maxParticipants || 0)}
                 onPressToCourse={readonly ? undefined : () => navigate(`/single-course/${subcourse.id}`)}
-                showSchoolclass
+                showSchoolclass={(subcourse as any)?.isInstructor || (subcourse as any)?.isParticipant ? true : false}
```
Jonasdoubleyou, line 47: ? true : false could be ?? false

Jonasdoubleyou (approved):

src/pages/subcourse/AllSubcourses.tsx
LomyW: `??` doesn't work with `||` , or?

Jonasdoubleyou: Yeah but isInstructor and isParticipant are already booleans, so we must only handle the null/undefined case ...

src/Utility.ts
```diff
@@ -130,9 +130,12 @@ export const getTrafficStatus: (participants: number, maxParticipants: number) =
...
-    if (!subcourse.published && subcourse.course.courseState === Course_Coursestate_Enum.Submitted) return i18next.t('single.global.courseState.submitted');
-    if (!subcourse.published && subcourse.course.courseState === Course_Coursestate_Enum.Allowed) return i18next.t('single.global.courseState.draft');
+    if (!subcourse.published) {
+        if (subcourse.course.courseState === Course_Coursestate_Enum.Created) return i18next.t('single.global.courseState.draft');
+        if (subcourse.course.courseState === Course_Coursestate_Enum.Submitted) return i18next.t('single.global.courseState.submitted');
+        if (subcourse.course.courseState === Course_Coursestate_Enum.Allowed) return i18next.t('single.global.courseState.draft');
+        if (subcourse.course.courseState === Course_Coursestate_Enum.Denied) return i18next.t('single.global.courseState.denied');
+    }
     if (subcourse.published) return i18next.t('single.global.courseState.publish');
     if (subcourse.cancelled) return i18next.t('single.global.courseState.cancelled');
```
Jonasdoubleyou, line 140: While looking at this, I'm not sure line 137 is actually reachable ... A cancelled course should also still be published, so it should hit 136 ... Could you check this in a followup task?

Jonasdoubleyou (approved):

PR #218, opened 2023-03-29
New match cards

LomyW:


src/pages/pupil/Matching.tsx
```diff
@@ -35,10 +36,12 @@ const query: DocumentNode = gql`
                         name
                     }
                     pupil {
+                        isPupil
                         schooltype
                         grade
                     }
                     student {
+                        isStudent
```
Jonasdoubleyou, line 44: Why do you need isPupil and isStudent here? These are very historic attributes, and they basically mean "wants to be a tutor" and "wants to be a tutee". There are students where isStudent is false and pupils where isPupil is false

Jonasdoubleyou: I think you rather want to use useUserType() for this, no?

src/pages/student/MatchingStudent.tsx
```diff
@@ -33,6 +35,7 @@ const query = gql`
                         name
                     }
                     pupil {
+                        isPupil
```
Jonasdoubleyou, line 38: isPupil?

Jonasdoubleyou (approved):

src/widgets/LearningPartner.tsx
```diff
@@ -1,100 +1,79 @@
...
-                    )}
-
-                    {button && (
-                        <Column mt={space['1']} space={space['1']}>
-                            {meetingId && (
-                                <Button space={2} onPress={() => window.open(`https://meet.jit.si/CoronaSchool-${meetingId}`, '_blank')}>
-                                    {t('matching.shared.videochat')}
-                                </Button>
-                            )}
-                            {contactMail && <Button onPress={() => (window.location.href = `mailto:${contactMail}`)}>{t('matching.shared.contact')}</Button>}
```
Jonasdoubleyou, line 91: Sorry, totally overlooked this: For now contacting via E-Mail and the Jitsi link are "better than nothing" till we have the Chat ... So maybe we can add them temporarily to the Match Card and comment out the buttons without functionality? If we would chat and Zoom next week I'd be fine to block main, but I guess It'll gonna take a while longer.

PR #219, opened 2023-03-29
Match detail page

LomyW:


src/pages/SingleMatch.tsx
```diff
@@ -0,0 +1,183 @@
...
+    );
+
+    useEffect(() => {
+        if (dissolveData?.matchDissolve && !toastShown) {
+            setToastShown(true);
+            toast.show({ description: t('matching.shared.dissolved'), placement: 'top' });
+        }
+    }, [dissolveData?.matchDissolve, toast, toastShown]);
+
+    const federalState = useMemo(() => {
```
Jonasdoubleyou, line 101: Shouldn't this be the other way round, so the pupil sees the state of the student and the student sees the state of the pupil? Either way, I would prefer to move this logic into the <MatchPartner /> components below, like `{userType === 'student' && <MatchPartner federalState={data.match.pupil.state} />`, and the  "Other handling" would then happen in the MatchPartner component

src/pages/SingleMatch.tsx
```diff
@@ -0,0 +1,183 @@
...
+    // const tabs: Tab[] = [
+    //     {
+    //         title: 'Termine',
+    //         content: <Text>Termine</Text>,
+    //     },
+    // ];
+
+    return (
+        <WithNavigation headerTitle={''} showBack headerLeft={<NotificationAlert />}>
+            {loading ? (
```
Jonasdoubleyou, line 120: As a small trick, instead of `loading` you could use `(loading || !data)` here, then `data` will be non nullable in the else branch, then one can omit the null handling in a few places

src/pages/SingleMatch.tsx
```diff
@@ -0,0 +1,183 @@
...
+                <CenterLoadingSpinner />
+            ) : (
+                !error &&
+                !loading && (
```
Jonasdoubleyou, line 124: `!loading` is redundant with the ternary condition

src/pages/match/MatchPartner.tsx
```diff
@@ -0,0 +1,62 @@
...
+type MatchPartnerProps = {
+    name: string;
+    schooltype?: string;
+    grade?: string;
+    federalState: string;
+    subjects: Subject[];
+    aboutMe: string;
+    isPupil: boolean;
+};
+const MatchPartner: React.FC<MatchPartnerProps> = ({ name, schooltype, grade, federalState, subjects, aboutMe, isPupil }) => {
```
Jonasdoubleyou, line 18: As an idea, would it make sense to pass a `matchPartner` attribute here instead of "flattening" all the properties? Like `<MatchPartner partner={match.pupil} isPupil />`? That would make the usage much more simple (and less flexible, but I doubt we ever want to override single properties of a student)

Jonasdoubleyou (commented):
Nice! (some minor comments)

src/pages/match/MatchPartner.tsx
```diff
@@ -0,0 +1,66 @@
...
+    }, [partner.state]);
+
+    const school = useMemo(() => {
+        if ('schooltype' in partner) return partner.schooltype !== Pupil_Schooltype_Enum.Other && partner.schooltype;
+        return undefined;
+    }, [partner]);
+
+    const matchPartnerInfos = useMemo(() => {
+        let strings: string[] = [];
+        school && strings.push(school);
```
Jonasdoubleyou, line 29: if (school) ...
if (... && grade) ...
if (state) ...

would be more consistent

Jonasdoubleyou (approved):
(some final touchups, looks good!)

PR #222, opened 2023-03-30
feat: left chat page

LucasFalkowsky:
A page that's reached after leaving a chat was implemented.
It can be reached on ./left-chat

### DIFF
--- a/src/pages/chat/LeftChat.tsx
+++ b/src/pages/chat/LeftChat.tsx
@@ -1,0 +1,42 @@
+import { Box, Button, Heading, Stack, Text, useBreakpointValue, useTheme, View } from 'native-base';
+import { useTranslation } from 'react-i18next';
+import { useNavigate } from 'react-router-dom';
+import PartyIcon from '../../assets/icons/lernfair/lf-party.svg';
+
+const LeftChat: React.FC = () => {
+    const width = useBreakpointValue({
+        base: '100%',
+        lg: '90%',
+    });
+    const buttonWidth = useBreakpointValue({
+        base: '100%',
+        md: '300px',
+    });
+    const { t } = useTranslation();
+    const navigate = useNavigate();
+
+    const { space } = useTheme();
+
+    return (
+        <View position="fixed" top="0" left="0" right="0" w="100vw" h="100vh" background="primary.900">
+            <Stack w={width} h="inherit" padding={'24px'} flex={1} space={space['1']} direction="column" justifyContent={'center'}>
+                <Box alignSelf={'center'}>
+                    <PartyIcon />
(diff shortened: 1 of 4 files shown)


src/pages/chat/LeftChat.tsx
```diff
@@ -0,0 +1,42 @@
...
+                </Box>
+                <Stack space={space['1']} direction={'column'}>
+                    <Heading fontWeight="700" lineHeight="md" fontSize={'lg'} color={'white'} textAlign={'center'}>
+                        {t('chat.course.leftChat.title')}
+                    </Heading>
+                    <Text fontWeight="normal" fontSize={'xs'} color={'white'} textAlign={'center'}>
+                        {t('chat.course.leftChat.subtitle')}
+                    </Text>
+                </Stack>
+                <Button bgColor={'yellow.400'} alignSelf={'center'} width={buttonWidth} onPress={() => navigate('/')}>
```
LomyW, line 34: you don't have to set a bgColor here, because there a pre-defined `variants` of buttons. 🙂 
Only need prop `variant` if you don't want a yellow primary button (like `outline`)

src/pages/chat/LeftChat.tsx
```diff
@@ -0,0 +1,42 @@
...
+        md: '300px',
+    });
+    const { t } = useTranslation();
+    const navigate = useNavigate();
+
+    const { space } = useTheme();
+
+    return (
+        <View position="fixed" top="0" left="0" right="0" w="100vw" h="100vh" background="primary.900">
+            <Stack w={width} h="inherit" padding={'24px'} flex={1} space={space['1']} direction="column" justifyContent={'center'}>
```
LomyW, line 22: props like `padding` have not to be wrapped into a `{ }` , should work without

src/pages/chat/LeftChat.tsx
```diff
@@ -0,0 +1,42 @@
+import { Box, Button, Heading, Stack, Text, useBreakpointValue, useTheme, View } from 'native-base';
+import { useTranslation } from 'react-i18next';
+import { useNavigate } from 'react-router-dom';
+import PartyIcon from '../../assets/icons/lernfair/lf-party.svg';
+
+const LeftChat: React.FC = () => {
```
LomyW, line 6: we should pass a prop wheter we came from a `course` or `oneOnOne` videochat

LomyW (changes requested):
well done! ✨

src/pages/chat/LeftChat.tsx
LucasFalkowsky: I deleted the color prop 👍

PR #226, opened 2023-04-02
Move Match buttons to Match Detail Page

Jonasdoubleyou:


Jonasdoubleyou:
Hey @LomyW, I moved the buttons from the match card to the match detail page. I think that better fits the new design (the match card looked weird on mobile).

PR #236, opened 2023-04-05
Lazy Retry of Chunk Load

Jonasdoubleyou:
This should reduce the number of times the app stops working as loading a module asynchronously fails to fetch (i.e. due to unreliable network connections).  As a next step, we could show a dedicated popup instead of the current 'an issue occured' if this is still a problem even with retries


src/lazy.ts
```diff
@@ -0,0 +1,29 @@
...
+    if (options.prefetch) {
+        // Already load the module now, but only use it if the component is actually rendered
+        // Ignore promise rejections (instead of leaving them unhandled)
+        loadModule().catch(error => {
+            log('Lazy Retry', `Failed to load module in prefetch ${error.message}`);
+        });
+    }
+
+    async function retry() {
+        for (let count = 1; count < RETRY_COUNT; count++) {
```
Leo310, line 16: Technically should be `for (let count = 1; count <= RETRY_COUNT; count++) {` or it would only be retried twice with a RETRY_COUNT of 3.

src/lazy.ts
```diff
@@ -0,0 +1,29 @@
+import { ComponentType, LazyExoticComponent, lazy } from 'react';
+import { log } from './log';
+
+const RETRY_COUNT = 3;
+
+export function lazyWithRetry<Component extends ComponentType<any>>(loadModule: () => Promise<{ default: Component }>, options: { prefetch?: boolean } = {}): LazyExoticComponent<Component> {
+    if (options.prefetch) {
```
Leo310, line 7: When prefetch is true isn't the Module imported twice? Once here and then later dynamically with lazy()?
What is the intention behind prefetch?

src/Navigator.tsx
```diff
@@ -9,12 +9,10 @@ import Welcome from './pages/Welcome';
...
 import { RequireAuth } from './User';
 import FullPageModal from './widgets/FullPageModal';
+import { lazyWithRetry } from './lazy';
 
 // All other pages load lazy:
-const NavigatorLazy = lazy(() => import('./NavigatorLazy'));
-
-// But as after login the user will visit the dashboard anyways, let's start loading it already
-import('./NavigatorLazy');
+const NavigatorLazy = lazyWithRetry(() => import('./NavigatorLazy'), { prefetch: true });
```
Leo310, line 15: In total, we have 2 bundles right? The main one loaded on start-up and the second bundle which includes the NavigatorLazy with every other route. 
Why aren't we lazy-loading the different routes? Wouldnt that improve loading speeds even more?

src/lazy.ts
Jonasdoubleyou: The third retry is the one in line 24

src/lazy.ts
Jonasdoubleyou: Yes, however 'importing twice' means 'loaded once'. Once the module was loaded successfully, importing it again will just return a reference to the same module from cache.

When you open the user app (and turn off caching in the user app, and turn on network throttling to better see the effects) you'll see that once the page is loaded, it already starts fetching the next chunk. If you click login, the page is already loaded. If you click register the loading spinner will show till the next chunk arrives. 

We prefetch so that while the user logs in, already the rest of the app is loaded so that it's instantly there when you click on login.

src/lazy.ts
Leo310: Ah yes I see

src/Navigator.tsx
Jonasdoubleyou: There is a third one for the icons. 

Not necessarily. Currently the app will load some time when opened for the first time (and when we release a new version) and interacting with the app will be smooth as only data needs to be fetched.

So it doesn't make sense to split this up into each and every page, but it would definetly make sense to split out rarely used pages such as the registration.

Leo310: Ok makes sense

Leo310 (approved):
:+1:

PR #243, opened 2023-04-12
Video Chat

LucasFalkowsky:
### Implementation of the Zoom video chat feature


src/components/ZoomMeeting.tsx
```diff
@@ -0,0 +1,88 @@
...
+    ZoomMtg.prepareWebSDK();
+    ZoomMtg.i18n.load('de-DE');
+    ZoomMtg.i18n.reload('de-DE');
+
+    const credentials = {
+        authEndpoint: '',
+        sdkKey: 'oy00hCKEQvKyxcR49FzEyw',
+        password: '',
+        meetingNumber: id!,
+        signature: data?.zoomSDKJWT || '',
```
Jonasdoubleyou, line 51: Can data be null? If yes, is an empty string valid as a signature?

src/components/ZoomMeeting.tsx
```diff
@@ -0,0 +1,88 @@
...
+    zoomSDKJWT (meetingId: $meetingId, role: $role)
+}`);
+
+const ZoomMeeting: React.FC = () => {
+    const { id, type } = useParams();
+
+    window.onpopstate = (e) => {
+        e.preventDefault();
+        document.getElementById('zmmtg-root')!.style.display = 'none';
+        window.location.reload();
```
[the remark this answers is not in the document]
LucasFalkowsky: because zoom does not stop the meeting when returning back from the zoom meeting page. The video camera and microphone will still be turned on and the meeting would be up and running.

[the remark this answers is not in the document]
Jonasdoubleyou: So we refresh the whole page, drop all session state, and interrupt the Zoom video streams just to exit the Zoom Meeting? Doesn't the Zoom API offer a method to exit more gracefully? This doesn't sound right to me.

Also shouldn't this then be in the cleanup of a global effect instead of registering an event listener on window?

src/config.ts
```diff
@@ -10,3 +10,5 @@ export const WEBSOCKET_URL = window.liveConfig?.RUNTIME_WEBSOCKET_URL ?? process
 
 export const DEACTIVATE_PUPIL_MATCH_REQUESTS =
     window.liveConfig?.RUNTIME_DEACTIVATE_PUPIL_MATCH_REQUESTS ?? process.env.REACT_APP_DEACTIVATE_PUPIL_MATCH_REQUESTS;
+
+export const ZOOM_MEETING_SDK_KEY = 'oy00hCKEQvKyxcR49FzEyw';
```
Jonasdoubleyou, line 14: Move to ENV?

src/components/ZoomMeeting.tsx
```diff
@@ -0,0 +1,110 @@
+// eslint-disable-next-line lernfair-app-linter/typed-gql
+import { gql, useQuery } from '@apollo/client';
```
Jonasdoubleyou, line 2: (You could also use the gql from 'src/gql' which is typesafe)

src/components/ZoomMeeting.tsx
```diff
@@ -0,0 +1,110 @@
...
+            email
+        }
+    }
+    zoomZAK
+    zoomSDKJWT (meetingId: $meetingId, role: $role)
+}`);
+
+const ZoomMeeting: React.FC = () => {
+    const { id, type } = useParams();
+    const appointmentId = parseInt(id!) || 0;
```
Jonasdoubleyou, line 37: If we expect an id in the params, we should fail if it does not exist, i.e. `if (!appointmentId) throw new Error('Missing id in the parameters of ZoomMeeting');`

src/components/ZoomMeeting.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+    useEffect(() => {
+        const handlePopState = () => {
+            window.location.reload();
```
Jonasdoubleyou, line 41: So Zoom has no other mechanism to end the meeting?

src/components/ZoomMeeting.tsx
```diff
@@ -0,0 +1,110 @@
...
+                },
+                error: (error: any) => {
+                    console.log(error);
+                },
+            });
+        },
+        error: (error: any) => {
+            console.log(error);
+        },
+    });
```
Jonasdoubleyou, line 105: The whole Zoom setup should also be an effect?

src/components/ZoomMeeting.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+    return <div className="App"></div>;
```
Jonasdoubleyou, line 107: What's the purpose of this div?

src/components/appointment/AppointmentDetail.tsx
```diff
@@ -8,6 +8,7 @@ import Avatars from './Avatars';
 import Description from './Description';
 import Buttons from './Buttons';
 import { DateTime } from 'luxon';
+// eslint-disable-next-line lernfair-app-linter/typed-gql
 import { gql, useMutation, useQuery } from '@apollo/client';
```
Jonasdoubleyou, line 12: (same here)

src/components/ZoomMeeting.tsx
```diff
@@ -0,0 +1,88 @@
...
+    console.log(credentials);
+
+    ZoomMtg.init({
+        leaveUrl: credentials.leaveUrl,
+        success: () => {
+            ZoomMtg.join({
+                signature: credentials.signature,
+                sdkKey: credentials.sdkKey,
+                meetingNumber: credentials.meetingNumber,
+                passWord: '',
```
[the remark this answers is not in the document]
timidak: It's really passWord 😄

src/components/ZoomMeeting.tsx
timidak: https://devforum.zoom.us/t/unmount-zoom-lib-in-react-after-leaving-page/54853

Nope and the Router "workaround" is not working.

src/widgets/appointment/AppointmentList.tsx
```diff
@@ -152,12 +153,20 @@ const AppointmentList: React.FC<Props> = ({
...
                     />
                 </Box>
             </Box>
         );
     };
 
+    const [_, setRefresh] = React.useState(0);
+
+    useInterval(() => {
+        setRefresh(new Date().getTime());
```
Jonasdoubleyou, line 167: What do we refresh here? The Apollo Client also has mechanisms to refresh query results.

Jonasdoubleyou (approved):

src/widgets/appointment/AppointmentList.tsx
timidak: I know, but we're just refreshing the UI, because an appointment gets an UI update (with a button to join the meeting) when an appointment is about to begin. This functionality was either broken, or wasn't working before. 
The data we get from the backend isn't stale at this moment, but the UI has to change.

timidak:
I'd merge the branch into the appointments feature branch, could you then build some kind of feature deployment, so that we can test the features heavily together, after the backend was merged, @Jonasdoubleyou?

PR #247, opened 2023-04-18
Use wannabe roles to show onboarding pages

Jonasdoubleyou:
With the changes made in https://github.com/corona-school/backend/pull/578, we can now drop additional requests when navigating between tabs and can rely on the user roles present in the session


LomyW (approved):
nice! 😍 🎉

PR #257, opened 2023-05-06
Fix lecture management

Jonasdoubleyou:
When we only show a subset of lectures, the index is not accurate


Jonasdoubleyou:
@LomyW does it make sense to bring this in, or do you rework the course creation anyways with the appointments changes?

PR #258, opened 2023-05-06
refactor: retire loginLegacy

dhenkel92:
To make the application more secure we want to finally retire the old token login mechanism. We'll keep both query parameter for now.


Jonasdoubleyou (approved):
nice!

PR #259, opened 2023-05-09
Text/misc

Maascha:


src/widgets/ImportantInformation.tsx
```diff
@@ -183,11 +186,23 @@ const ImportantInformation: React.FC<Props> = ({ variant }) => {
 
         // -------- Pupil Screening --------
         if (pupil?.screenings.some((s) => !s.invalidated && s.status === 'pending')) {
+            const pupil_url =
+                process.env.REACT_APP_PUPIL_SCREENING_URL +
+                '?first_name=' +
+                encodeURIComponent(data?.me?.firstname as string) +
```
Jonasdoubleyou, line 192: I'd prefer ?? "" instead of as any as in case the property is not set we would pass "null" or "undefined"

Jonasdoubleyou (dismissed):

Jonasdoubleyou (approved):

PR #261, opened 2023-05-11
feat: chat context and hook

LomyW:
- [x] context to create session
- [x] added useChat hook

### DIFF
--- a/src/context/ChatContext.tsx
+++ b/src/context/ChatContext.tsx
@@ -1,9 +1,52 @@
-import React from 'react';
+import { ReactNode, createContext, useContext, useEffect, useState } from 'react';
+import Talk from 'talkjs';
+import useApollo from '../hooks/useApollo';
 
-const ChatContext = () => {
-    // TODO create session
-    // * getOrCreateConversation
-    return <div>ChatContext</div>;
+type IChatContext = {
+    session: Talk.Session | null;
+    talkLoaded: boolean;
+};
+const ChatContext = createContext<IChatContext>({
+    session: null,
+    talkLoaded: false,
+});
+
+export const LFChatProvider: React.FC<{ children: ReactNode }> = ({ children }) => {
+    const [session, setSession] = useState<Talk.Session | null>(null);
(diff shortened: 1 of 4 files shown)


src/context/ChatContext.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+    const me = {
+        id: '123',
+        name: 'Lomy',
+        role: 'student',
+        email: 'salome.wick@typedigital.de',
+    };
+
+    useEffect(() => {
+        Talk.ready.then(() => markTalkLoaded(true));
```
Jonasdoubleyou, line 31: Registered multiple times, should be in a separate useEffect with no dependencies

src/context/ChatContext.tsx
```diff
@@ -0,0 +1,52 @@
...
+                appId: 't5NarFaG',
+                me: currentUser,
+                // signature:
+            });
+            setSession(session);
+            return () => session.destroy();
+        }
+    }, [talkLoaded]);
+
+    return <ChatContext.Provider value={{ session, talkLoaded }}>{children}</ChatContext.Provider>;
```
Jonasdoubleyou, line 46: I think we can prevent quite some rerenders by moving the context value into a useMemo

src/components/WithNavigation.tsx
```diff
@@ -58,6 +50,16 @@ const WithNavigation: React.FC<Props> = ({
         lg: space['1'],
     });
 
+    const { t } = useTranslation();
+
+    const navItems: NavigationItems = {
+        start: { label: t('navigation.label.start'), icon: LFHomeIcon },
+        appointments: { label: t('navigation.label.appointments'), icon: LFAppointmentIcon },
+        chat: { label: t('navigation.label.chat'), icon: LFChatIcon },
```
Jonasdoubleyou, line 58: if you want to bring this to main, please remove not yet finished pages from the nav bar

LomyW: switched the base branch 😄

Jonasdoubleyou (approved):

PR #262, opened 2023-05-11
refactor: remove token

dhenkel92:
This is a cleanup PR to finally remove the 'token' query parameter


Jonasdoubleyou (changes requested):
(blocked till mails are all adapted)

PR #276, opened 2023-06-12
feat: contact one participant

LomyW:
- add button to create chat with a participant as a instructor
- only show `contact instructor button` if contact is allowed
- new button order


src/pages/pupil/single-course/PupilCourseButtons.tsx
```diff
@@ -2,19 +2,33 @@ import { ApolloQueryResult } from '@apollo/client';
...
+    participantsCount: number;
+    maxParticipants: number;
+    isParticipant: boolean;
+    isOnWaitingList: boolean;
+    canContactInstructor: { allowed: boolean };
+    conversationId?: string | null | undefined;
+    allowChatContactProspects?: boolean;
+    allowChatContactParticipants?: boolean;
+    groupChatType: Chat_Type;
+};
```
timidak, line 30: What about something like:
```suggestion
type SubcourseOfPupil = Subcourse & {
    groupChatType: Chat_Type;
};
```
This approach helps you maintain the types better, if `Subcourse` changes, you're just using the new type instead of having to extend it.

src/pages/pupil/single-course/PupilCourseButtons.tsx
```diff
@@ -125,9 +134,14 @@ const PupilCourseButtons: React.FC<ActionButtonProps> = ({
                 )}
 
                 {subcourse.isParticipant && (
-                    <Button onPress={() => setSignOutModal(true)} isDisabled={loadingSubcourseLeft}>
-                        {t('single.actions.leaveSubcourse')}
-                    </Button>
+                    <OpenSubcourseChat
```
timidak, line 137: This is a `button`, right? So add the suffix button to the Component please 🙂

src/pages/pupil/single-course/PupilCourseButtons.tsx
timidak: Why was this resolved?

src/pages/subcourse/ParticipantRow.tsx
```diff
@@ -1,26 +1,35 @@
...
                 <Text>
                     {participant.schooltype && `${getSchoolTypeKey(participant.schooltype)}, `}
                     {participant.grade}
                 </Text>
-            </Column>
-        </Row>
+            </VStack>
+            <Spacer />
+            {isInstructor && contactParticipant && (
+                <Button variant="outlinelight" ml={space['3']} onPress={() => contactParticipant(`pupil/${participant.id}`)}>
```
timidak, line 28: Do we have a util for adding `pupil/` or `student/` before an id? Otherwise changes would lead to massive refactorings all over the place.

timidak (commented):
Renaming the button is mandatory for an approve. 😄

src/pages/pupil/single-course/PupilCourseButtons.tsx
LomyW: renamed in another PR to `OpenCourseChatButton`

timidak (approved):
Nice work!

PR #290, opened 2023-06-26
fix: refetch data after decline or cancel appointment

LomyW:
- add a `refetching query` if an appointment was dissolved or cancelled


timidak:
Is this then working for Subcourses, as well? @LomyW

timidak (approved):
LGTM 👍🏻

PR #293, opened 2023-06-26
fix: save meeting report

LomyW:


src/pages/chat/LeftVideoChat.tsx
```diff
@@ -54,6 +54,14 @@ const LeftVideoChat: React.FC = () => {
         })();
     }, []);
 
+    useEffect(() => {
+        (async () => {
+            if (isOrganizer) {
+                await appointmentSaveMeetingReport({ variables: { appointmentId: idAsInt } });
+            }
+        })();
+    }, [isOrganizer]);
```
timidak, line 63: listen to the loading state of the appointment

src/pages/chat/LeftVideoChat.tsx
```diff
@@ -54,6 +54,14 @@ const LeftVideoChat: React.FC = () => {
         })();
     }, []);
 
+    useEffect(() => {
+        (async () => {
```
timidak, line 58: add an if to check wether it's loading and the data is present.

timidak (approved):
We should test this after it has been deployed 👍🏻

PR #304, opened 2023-07-10
fix: overlapping by updating zoom websdk to 2.14.0

LomyW:
**Description**
Update zoom websdk to version 2.14.0.
Fixes the overlap problems we had. 😄


public/index.html
```diff
@@ -17,8 +17,8 @@
         <meta name="msapplication-config" content="/favicons/browserconfig.xml" />
         <meta name="theme-color" content="#ffffff" />
 
-        <link type="text/css" rel="preload" href="https://source.zoom.us/2.11.5/css/bootstrap.css" />
-        <link type="text/css" rel="preload" href="https://source.zoom.us/2.11.5/css/react-select.css" />
+        <link type="text/css" rel="stylesheet" href="https://source.zoom.us/2.14.0/css/bootstrap.css" />
```
timidak, line 20: Can you try if the preload has any effect on that bug?

LomyW: with preload the overlap occurs

timidak (approved):

PR #305, opened 2023-07-10
755 change course tile layout (first two subissues)

Fabian5150:
Fixes first 2 subissues of https://github.com/corona-school/project-user/issues/755
Requesting already, because I had some very strange merge issues this earlier, now it works. The fixes of the other subissues should follow soon.


src/widgets/AppointmentCard.tsx
```diff
@@ -240,26 +243,22 @@ const AppointmentCard: React.FC<Props> = ({
...
-                                                      })
-                                                    : t('single.card.oneLecture', {
-                                                          date: date.toFormat('dd.MM.yyyy'),
-                                                          time: date.toFormat('HH:mm'),
-                                                          count: countCourse,
-                                                      })}
+                                                {/* TODO: Replace dateFirstLecture here with dateNextLecture. Data for that must be fetched
+                                                in all parent components of this one first and passed down as prop. See issue #755 */}
+                                                {t('single.card.dateLectures', {
+                                                    weekday: t('single.global.weekdays', { returnObjects: true })[dateFirstLecture.weekday - 1],
```
Jonasdoubleyou, line 253: Hm, maybe we could also use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat `dateStyle: "full"` here instead of maintaining our own localization?

Jonasdoubleyou (dismissed):

src/widgets/AppointmentCard.tsx
Fabian5150: I'll try that out. :)

Fabian5150: Couldn't used that specific one, because it works with Date type and we have 'DateTime' types here.
Instead used https://moment.github.io/luxon/docs/class/src/datetime.js~DateTime.html#instance-method-toLocaleString
which basically does the same thing on the 'DateTime' type.

Jonasdoubleyou (approved):
nice!

PR #308, opened 2023-07-13
fix: mobile view

LomyW:
### Description
We had the problem that the width of the user-app was "to big" on mobile.

### What was done?
- [x] added style to fix the min width for `body`


public/index.html
```diff
@@ -35,3 +35,10 @@
         <!-- <div id="aria-notify-area"></div> -->
     </body>
 </html>
+
+<style>
```
Jonasdoubleyou, line 39: Style tag outside <html> is invalid HTML? Also why can't we place this in one of the global .css files?

Jonasdoubleyou (changes requested):

Jonasdoubleyou (approved):

PR #310, opened 2023-07-14
fix: update contacts immediately

LomyW:
### Description
Contacts didn't update immediately. 

### What was done?
- [x] added refetch on open modal


src/components/chat/ContactList.tsx
```diff
@@ -84,6 +85,11 @@ const ContactList: React.FC<NewChatProps> = ({ onClose, setChatId }) => {
             toast.show({ description: t('chat.chatError'), placement: 'top' });
         }
     };
+
+    useEffect(() => {
+        refetch();
```
Jonasdoubleyou, line 90: Why exactly do we need to fetch twice when rendering the component the first time?

LomyW: Honestly, I'm not quite sure why myself. It didn't update the data (think that's modal). That's how it worked! 😄 👀

Jonasdoubleyou (approved):

PR #314, opened 2023-07-19
fix: courses without lectures

dhenkel92:
This can happen if a user deletes all lectures after the course has been created. The problem is that the course is not editable anymore and the user-app just crashes.

This PR will make sure that this doesn't happen anymore.

https://lern-fair.slack.com/archives/C03E45MES4X/p1689785056516519

### DIFF
--- a/src/pages/course-creation/CourseAppointments.tsx
+++ b/src/pages/course-creation/CourseAppointments.tsx
@@ -58,7 +58,6 @@ const CourseAppointments: React.FC<Props> = ({ next, back, isEditing, appointmen
         return convertedAppointments;
     };
     const canGoFurther = () => {
-        if (isEditing) return appointments.length === 0 ? true : false;
         return allAppointmentsToShow.length === 0 ? true : false;
     };
     const getAllAppointmentsToShow = () => {
--- a/src/modals/Unsplash.tsx
+++ b/src/modals/Unsplash.tsx
@@ -48,7 +48,7 @@ const Unsplash: React.FC<Props> = ({ showUnsplashModal, onPhotoSelected, onClose
             );
             const res = await data.json();
             setTotalPages(res.total_pages);
-            setPhotos(res.results);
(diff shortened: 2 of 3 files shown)


src/pages/course-creation/CourseAppointments.tsx
```diff
@@ -58,7 +58,6 @@ const CourseAppointments: React.FC<Props> = ({ next, back, isEditing, appointmen
         return convertedAppointments;
     };
     const canGoFurther = () => {
-        if (isEditing) return appointments.length === 0 ? true : false;
```
dhenkel92, line 61: I have no idea what was the idea behind this, but it prevents us from editing a course that doesn't have lectures.

Jonasdoubleyou: Hm, that state should not be reachable. All courses should at least have one lecture.

Jonasdoubleyou (approved):

src/pages/course-creation/CourseAppointments.tsx
dhenkel92: But then we also have to communicate to the user how to replace all lectures.

Jonasdoubleyou: Fair, the UI for course creation / editing was always messy, I wanted to revisit that problem once we have appointments. I guess if we allow this again then somewhen we will have the opposite bugs in the backend, were we rely on this assumption. But fine for me if it fixes the immediate problem

Jonasdoubleyou: Hm, at least https://github.com/corona-school/backend/blob/master/common/courses/states.ts looks as if we at least handle this in the majority of cases

dhenkel92: That's true. Prisma helps a lot here.

PR #316, opened 2023-07-20
fix: clear appointments on course creation

LomyW:
### Description
We have the problem that if the course creation is not completed because, for example, another tab is clicked, the appointments are not cleaned up.

### Solved
Every time we open the `course creation` we clear the `appointments to be created`


src/pages/CreateCourse.tsx
```diff
@@ -818,6 +819,10 @@ const CreateCourse: React.FC = () => {
         setCurrentIndex(index);
     }, []);
 
+    useEffect(() => {
```
LomyW, line 822: @timidak that's the only thing I could think of to solve it. Not convinced myself. Do you have another idea?

timidak: You could add your code `setAppoiontmensToBeCreated([])` in the cleanup function of the effect above, this should work I guess.

timidak: so basically you should add the following to the effect above:
```ts
return () => setAppoiontmensToBeCreated([])
```

src/pages/CreateCourse.tsx
```diff
@@ -496,6 +497,7 @@ const CreateCourse: React.FC = () => {
             // errors === undefined
             if (appointmentsRes.errors) {
                 errors.push('appointments');
+                setAppointmentsToBeCreated([]);
```
timidak, line 500: By moving this line before the if (but after the `createGroupAppointments`) you could also remove the line  `if (!appointmentsRes.errors) setAppointmentsToBeCreated([]);`.

src/pages/CreateCourse.tsx
```diff
@@ -289,6 +289,7 @@ const CreateCourse: React.FC = () => {
             documentTitle: 'Kurs erstellen',
         });
         // eslint-disable-next-line react-hooks/exhaustive-deps
+        return () => setAppointmentsToBeCreated([]);
```
timidak, line 292: That worked, right?

timidak (dismissed):
Just a suggestion, but nice work 👍🏻

src/pages/CreateCourse.tsx
LomyW: yes

timidak (approved):
LGTM 👍🏻

PR #320, opened 2023-08-07
feat: ad hoc meeting for matches

LomyW:


src/pages/SingleMatch.tsx
```diff
@@ -128,6 +139,15 @@ const SingleMatch = () => {
...
         setCreateAppointment(false);
     };
+
+    const startAdHocMeeting = async () => {
+        const meetingData = await createAdHocMeeting({ variables: { matchId: matchId } });
+        const appointmentId = meetingData && meetingData.data?.matchCreateAdHocMeeting.id;
+        const appointmentType = meetingData && meetingData.data?.matchCreateAdHocMeeting.appointmentType;
+        await refetch();
+        navigate(`/video-chat/${appointmentId}/${appointmentType}`);
+    };
```
timidak, line 149: useCallback? And also what if the appointmentType oder Id is undefined?
We should throw an error if so imho. This would then trigger the ErrorBoundary.

src/modals/AdHocMeetingModal.tsx
```diff
@@ -0,0 +1,47 @@
+import { Button, Modal, Row, Text, useBreakpointValue, useTheme } from 'native-base';
+import { useMemo, useState } from 'react';
+import { useUserType } from '../hooks/useApollo';
+import { useNavigate } from 'react-router-dom';
+import { useTranslation } from 'react-i18next';
+
+type AdHocMeetingModal = {
```
timidak, line 7: These are props so at least suffix with `Props`

src/modals/AdHocMeetingModal.tsx
```diff
@@ -0,0 +1,47 @@
...
+    showAdHocModal: boolean | undefined;
```
timidak, line 8: Why can this be undefined?

timidak (commented):
A few changes to be made 😄

PR #321, opened 2023-08-09
feature: install DD real-user-monitoring

dhenkel92:
There are a couple of environment variables that have to be set in staging/production:

* `DATADOG_API_KEY` & `DATADOG_SITE` to upload the source maps
* `RUNTIME_DD_ENV` to distinguish the environments within DataDog
* `RUNTIME_DD_CLIENT_TOKEN` & `RUNTIME_DD_APP_ID` to authenticate RUM 

Besides that we have to enable the `runtime-dyno-metadata` lab for all user-apps and backend environments.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,8 @@
     "devDependencies": {
         "@apollo/client": "^3.6.9",
         "@babel/core": "^7.16.0",
+        "@datadog/browser-rum": "https://gitpkg.now.sh/corona-school/browser-sdk/packages/rum?lern-fair-packages",
+        "@datadog/datadog-ci": "^2.18.0",
         "@graphql-codegen/cli": "2.15.0",
         "@graphql-codegen/client-preset": "^1.2.1",
         "@jonkoops/matomo-tracker-react": "^0.7.0",
@@ -121,7 +123,8 @@
         "linter": "eslint . --ext .tsx",
         "graphql-fetch-schema": "node scripts/fetch-graphql-schema.js",
         "graphql-watch": "graphql-codegen --config config/graphql-codegen.ts --watch",
-        "graphql-build": "graphql-codegen --config config/graphql-codegen.ts"
+        "graphql-build": "graphql-codegen --config config/graphql-codegen.ts",
+        "heroku:release": "./scripts/dd-upload-source-maps.sh"
     },
     "eslintConfig": {
         "extends": [
--- a/scripts/dd-upload-source-maps.sh
+++ b/scripts/dd-upload-source-maps.sh
@@ -1,0 +1,31 @@
+#! /usr/bin/env bash
+
+set -eo pipefail
+
+appVersion=${HEROKU_SLUG_COMMIT:-latest}
+env=${DD_ENV:-dev}
+serviceName=${SERVICE_NAME:-user-app}
+
+
+# RUM should not be active in our review apps
+if [ "$env" == "dev" ]; then
+	echo "Not uploading source maps in dev"
+	exit 0
+fi
+
+pathPrefix="http://localhost:3000"
+case "$env" in
+    "staging")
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -197,6 +199,7 @@
         ]
     },
     "dependencies": {
+        "@datadog/browser-rum": "https://gitpkg.now.sh/corona-school/browser-sdk/packages/rum?lern-fair-packages",
```
dhenkel92, line 202: Maybe there is a better way than using this external gitpkg.now.sh service?

scripts/dd-upload-source-maps.sh
```diff
@@ -0,0 +1,30 @@
...
+
+appVersion=${HEROKU_SLUG_COMMIT:-latest}
+env=${DD_ENV:-dev}
+serviceName=${SERVICE_NAME:-user-app}
+
+
+# RUM should not be active in our review apps
+if [ "$env" == "dev" ]; then
+	echo "Not uploading source maps in dev"
+	# exit 0
```
dhenkel92, line 13: Comment should be removed after testing.

scripts/dd-upload-source-maps.sh
```diff
@@ -0,0 +1,30 @@
+#! /usr/bin/env bash
```
dhenkel92, line 1: Unfortunately, we cannot upload the source map in CI, because the files generated during the heroku release might be different...

src/index.tsx
```diff
@@ -2,9 +2,36 @@ import React from 'react';
...
+    version: APP_VERSION,
+    sessionSampleRate: 100,
+    sessionReplaySampleRate: 100,
+    trackUserInteractions: true,
+    trackResources: true,
+    trackLongTasks: true,
+    traceSampleRate: 100,
+    defaultPrivacyLevel: 'mask',
+    // TODO: Add proper tracing urls
+    allowedTracingUrls: [() => true],
```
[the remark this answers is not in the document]
dhenkel92: @Jonasdoubleyou: 
> Does this exclude secret_token?

package.json
Jonasdoubleyou: As I said, we can npm publish the package

src/index.tsx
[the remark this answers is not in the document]
dhenkel92: This one if fall all the requests that the user-app is sending.
The idea is that you put a list of domains that should have the tracing information injected, so that it's not sent to everyone.

package.json
dhenkel92: How do we resolve this one then?
https://github.com/corona-school/browser-sdk/pull/3/files#diff-b4a410833047c5ceeddca578774d17239ba2e0380423402768be4fa9299af18bR15

Jonasdoubleyou: Publish multiple packages?

dhenkel92: But as far as I understand we can't use their @datadog package name, can't we?
Which would mean that we would have to replace all imports with our new package.

Jonasdoubleyou: https://github.com/npm/npm/issues/2943

Ah true :/

Jonasdoubleyou: We could git clone into .browser-sdk, then nom import a relative path

dhenkel92: Good idea and would definitely work.
But tbh. I don't know if I like it more. 😅  Submodules have their own pain.

dhenkel92: We are going to keep the current solution and I'll create a ticket to revisit the topic later.

Jonasdoubleyou, line 202: I think this can be a devDependency, we only need it during build time when building the bundle, not in the running webserver

Jonasdoubleyou (approved):

package.json
dhenkel92: Done

PR #322, opened 2023-08-09
fix: input problem on appointment creation on safari

LomyW:
### Description
We had the problem, that only a single character is accepted on controlled input.
So it seems that the problem occures when a select element is present with an input on the same page.
refers to https://github.com/corona-school/project-user/issues/853

### Solved
- [x] create custom select component

### DIFF
--- a/src/pages/create-appointment/AppointmentForm.tsx
+++ b/src/pages/create-appointment/AppointmentForm.tsx
@@ -1,4 +1,4 @@
-import { Box, FormControl, Select, Stack, TextArea, useBreakpointValue, VStack, WarningTwoIcon } from 'native-base';
+import { Box, FormControl, Stack, TextArea, useBreakpointValue, VStack, WarningTwoIcon } from 'native-base';
 import { useTranslation } from 'react-i18next';
 import DatePicker from '../../components/DatePicker';
 import { useCreateAppointment } from '../../context/AppointmentContext';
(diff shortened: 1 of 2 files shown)


src/pages/create-appointment/AppointmentForm.tsx
```diff
@@ -138,7 +139,7 @@ const AppointmentForm: React.FC<FormProps> = ({ errors, appointmentsCount, onSet
                     {/* DURATION */}
                     <FormControl isInvalid={'duration' in errors} width={inputWidth}>
                         <FormControl.Label>{t('appointment.create.durationLabel')}</FormControl.Label>
-                        <Select placeholder="Dauer der Unterrichtseinheit" onValueChange={(e) => handleDurationSelection(e)}>
+                        <Select placeholder="Dauer der Unterrichtseinheit" onValueChange={(duration: string) => handleDurationSelection(duration)}>
```
Jonasdoubleyou, line 142: How did this even work before? 😅

src/widgets/Select.tsx
```diff
@@ -0,0 +1,11 @@
+import { Select as NBSelect } from 'native-base';
+
+const Select = (props: any) => {
+    const isSafari = () => /Safari/.test(navigator.userAgent) && /Apple Computer/.test(navigator.vendor);
+
+    return <NBSelect {...props} selection={isSafari() ? 1 : null} />;
```
Jonasdoubleyou, line 6: Where does this hack come from? Could you add a comment, cause I do not understand how it works?

Jonasdoubleyou (approved):

src/widgets/Select.tsx
LomyW: So I found the same issue here: https://github.com/GeekyAnts/NativeBase/issues/5111
And the solution therefore is: https://github.com/spr-networks/super/blob/main/frontend/src/components/Select.web.js
I honestly cant explain to you how it works... 😅

PR #329, opened 2023-08-12
feat: Pupil Screening

Jonasdoubleyou:
Adds the possibility for Screeners to Log in to the User App, see pupils with disputed screenings (another screener did the screening but did not decide yet) as well as search for pupils (for example when a pupil booked a screening appointment). Screeners can then leave a screening comment, mark the screening as success or rejection, or deactivate the pupil's account.

----

Additionally fixes and documents stateful Storybooks.

----

To test the actual Screening, request a Screening for i.e. "Max Musterschüler":
```
mutation { 
	pupilCreateScreening(pupilId: 1)
}
```

When done, login with auth token "authtokenSC1" or "test+dev+sc1@lern-fair.de" and password "testtest" (changed it on staging, after the next deployment it'll be "test" again). Then search for pupils, with i.e. "test+dev+p1@lern-fair.de" in the searchbar.

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -28,7 +28,7 @@ export const decorators = [
             <BrowserRouter>
                 <NativeBaseProvider theme={Theme}>
                     <MockedProvider mocks={[]}>
-                        <Page />
+                        {Page()}
                     </MockedProvider>
                 </NativeBaseProvider>
             </BrowserRouter>
--- a/src/User.tsx
+++ b/src/User.tsx
@@ -43,12 +43,12 @@ export const SwitchUserType = ({
 
     if (user!.student) {
         if (studentComponent) return studentComponent;
-        else return <Navigate to="/dashboard" state={{ from: location }} replace />;
+        else return <Navigate to="/start" state={{ from: location }} replace />;
     } else if (user!.pupil) {
         if (pupilComponent) return pupilComponent;
-        else return <Navigate to="/dashboard" state={{ from: location }} replace />;
+        else return <Navigate to="/start" state={{ from: location }} replace />;
     } else {
         if (screenerComponent) return screenerComponent;
-        else return <>Huh?</>;
+        else return <Navigate to="/start" state={{ from: location }} replace />;
     }
 };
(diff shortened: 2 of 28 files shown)


Jonasdoubleyou:
ready for review, "verify against prod" fails as the related backend change has not yet reached production

.storybook/preview.tsx
```diff
@@ -28,7 +28,7 @@ export const decorators = [
             <BrowserRouter>
                 <NativeBaseProvider theme={Theme}>
                     <MockedProvider mocks={[]}>
-                        <Page />
+                        {Page()}
```
Jonasdoubleyou, line 31: The Page Component might need to access some global context (i.e. if the mocked `useState` is used), which is only set directly when the storybook is rendered, so in this decorator we also must render directly and not defer to the <Page /> component render.

src/User.tsx
```diff
@@ -15,15 +15,23 @@ export const RequireAuth = ({ children, isRetainPath }: { children: JSX.Element;
     if (sessionState === 'unknown' || !user) return <CenterLoadingSpinner />;
 
     if (sessionState === 'logged-in') {
-        if (user && !(user.pupil ?? user.student)!.verifiedAt) return <VerifyEmailModal email={user.email} />;
+        if (user && !user.screener && !(user.pupil ?? user.student)!.verifiedAt) return <VerifyEmailModal email={user.email} />;
```
Jonasdoubleyou, line 18: Screeners get their Account manually through Support, so the E-Mail should be correct ...

src/widgets/matching/MatchStudentCard.tsx
```diff
@@ -0,0 +1,70 @@
+import { Text, useTheme, Box, Pressable, useBreakpointValue, HStack, Center, VStack } from 'native-base';
+
+import { useTranslation } from 'react-i18next';
+
+import StudentAvatar from '../../assets/icons/lernfair/avatar_student_56.svg';
+import Tag from '../../components/Tag';
+import { MatchWithStudent } from '../../types';
+
+export function MatchStudentCard({ match }: { match: MatchWithStudent }) {
```
Jonasdoubleyou, line 9: This is pretty close to "LearningPartner", however LearningPartner depends on the current user to either show the pupil or student of a match. Instead of making that component more complicated I decided to carve out only the Student variant.

realmayus (approved):

PR #342, opened 2023-08-20
Fix safari scroll

realmayus:
This fixes an issue with Safari and perhaps other mobile browsers that stems from the use of the `vh` unit in CSS. In newer versions of Safari, a dynamically-sized navbar is used, which is why the `dvh` unit is preferred as it automatically represents this resize. The PR also gets rid of fixed positioning of the bottom navbar in favor of flexbox.

Other than that, this PR also changes some minor paddings to make the user-app look better on iOS and changes the min-width of the chat report modal which was unusable on mobile.
Closes https://github.com/corona-school/project-user/issues/816


Jonasdoubleyou (dismissed):

Jonasdoubleyou:
Sweet, at least for me this seems to look much better on Safari - especially when zooming in and out this fixes loads of UI bugs. Also for me on Firefox the Mobile Device Emulation now also works much better!

Jonasdoubleyou:
There is a "min-width: 1100px" on the "body" somewhere that still causes a scroll on Mobile Device Emulation

Jonasdoubleyou (approved):

PR #346, opened 2023-08-21
fix: Students need the Instructor Role before they can Create a Course

Jonasdoubleyou:
This is a workaround, the proper solution would be to somehow refresh the roles directly after the screening


Jonasdoubleyou:
c.f. https://github.com/corona-school/project-user/issues/852

LomyW (approved):
Nice! 😃

PR #347, opened 2023-08-21
feat: Remove issue reporting

Jonasdoubleyou:
Datadog tracks all navigations and errors anyways, so no need to do it "home grown" a second time.


src/log.ts
```diff
@@ -1,33 +1,14 @@
...
-const logBuffer = new LogBuffer(20);
-export const getLastLogs = () => logBuffer.read();
+import { datadogRum } from '@datadog/browser-rum';
 
 export function log(component: string, message: string, ...context: any[]) {
     const line = `[${component}] ${message}`;
     console.log(line, ...context);
-    logBuffer.write(line);
+
+    datadogRum.addAction(line);
```
dhenkel92, line 7: Let's see how this will work out. Unfortunately, we can only attach errors otherwise 🙄

dhenkel92 (approved):
😭 😍

PR #348, opened 2023-08-22
fix: notification preferences

LomyW:
refers to: https://github.com/corona-school/backend/pull/809
- [x] remove `parse` and `stringify` notification preferences

### DIFF
--- a/src/helper/notification-helper.ts
+++ b/src/helper/notification-helper.ts
@@ -119,7 +119,7 @@ const getAllPreferencesInCategorySetToValue = (
     return newPreferences;
 };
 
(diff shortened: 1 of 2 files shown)


src/helper/notification-helper.ts
```diff
@@ -119,7 +119,7 @@ const getAllPreferencesInCategorySetToValue = (
     return newPreferences;
 };
 
-const getPreferencesCopy = (preferences: NotificationPreferences): NotificationPreferences => JSON.parse(JSON.stringify(preferences));
+const getPreferencesCopy = (preferences: NotificationPreferences): NotificationPreferences => preferences;
```
Jonasdoubleyou, line 122: Where is this used, and why don't we need to deep clone anymore?

Jonasdoubleyou (approved):

PR #349, opened 2023-08-22
fix: add more prod domains for source map upload

dhenkel92:


Jonasdoubleyou:
related https://github.com/corona-school/project-user/issues/830

PR #356, opened 2023-08-31
fix: increase time to join meeting before

LomyW:
### Description
increase `join before minutes` to 4 hours instead of 30 minutes


Jonasdoubleyou (approved):
thanks!

PR #358, opened 2023-09-05
Fix friendly relative time

realmayus:
Fixes decision when a date is considered "tomorrow" (only checked relative time difference, not date of the month)

Issue: 
image


src/Utility.ts
```diff
@@ -20,7 +20,16 @@ export const toTimerString = (refDate: number, compareDate: number) => {
 
     if (days > 1) return `In ${days} Tagen`;
```
Jonasdoubleyou, line 21: Isn't this also off then? Can't we do

const days = date.getDate() - ref.getDate(); ?

realmayus: Hm yeah, but what about the case when comparing e.g. 2023-09-05 and 2023-10-06? That would only give us 1

Jonasdoubleyou: DateTime.now().startOf("day").diff( start.startOf("day"), "day")

PR #359, opened 2023-09-11
feat: Cooperations

Jonasdoubleyou:


Jonasdoubleyou:
c.f. https://user-app-feat-cooperati-zrskmc.herokuapp.com/registration/helper?cooperation=self

src/pages/Registration.tsx
```diff
@@ -99,51 +104,70 @@ const Registration: React.FC = () => {
...
+        )
+    );
+
+    const cooperationTag = new URL(window.location.toString()).searchParams.get('cooperation');
+
+    const cooperation = useMemo(() => {
+        const result = corpData?.cooperations?.find((it) => it.tag === cooperationTag);
+
+        if (cooperationTag && corpData && !result) {
+            throw new Error(`Invalid Cooperation Tag '${cooperationTag}'`);
```
dhenkel92, line 130: /nit
Should be fine for our fast DHL change, but I think we generally should try to avoid crashing hard

dhenkel92 (approved):
Tested the change in the review app. LGTM 👍

src/pages/Registration.tsx
Jonasdoubleyou: Hmmm, I think a misconfiguration of this should rather be detected by us (this shows the issue reporter) than being ignored (and thus we might report wrong numbers)

dhenkel92: That's true, but failing hard is different than explaining what happened. 
It's the first time the people are interacting with our page and they have no idea if it's down or if it's just a broken link.

Btw. should there be an option to disable the cooperation?

Jonasdoubleyou: Fair enough, I'll build something for that

PR #360, opened 2023-09-12
Fix: Entering appointment infos on safari

LomyW:
### Description
Select element takes focus from input element on react native web on Safari browser

### Done
- [x] Added a select specific for safari

### Tested
- [x] Safari
- [x] Firefox
- [x] Chrome


timidak:
@LomyW is this branch up to date with main? 
Seems like the checks are failing, because of the graphql validation.

LomyW:
> @LomyW is this branch up to date with main? Seems like the checks are failing, because of the graphql validation.

yes it is up to date with main.

dhenkel92 (dismissed):

src/widgets/SafariSelect.tsx
```diff
@@ -0,0 +1,11 @@
+import { Select as NBSelect } from 'native-base';
+
+const SafariSelect = (props: any) => {
```
timidak, line 3: `props` shouldn't be typed as any. Use the props the `NBSelect` uses.
Also, `SafariSelect` seems like it's only used for Safari?

src/widgets/SafariSelect.tsx
```diff
@@ -0,0 +1,11 @@
...
+    const isSafari = () => /Safari/.test(navigator.userAgent) && /Apple Computer/.test(navigator.vendor);
+
+    return <NBSelect {...props} selection={isSafari() ? 1 : null} />;
+};
+
+SafariSelect.Item = NBSelect.Item;
+export default SafariSelect;
+export { SafariSelect };
```
timidak, line 11: Why do we have to export statements? I'd say either try the named export or the default one.

src/widgets/SafariSelect.tsx
timidak: I would add some additional documentation, why we had do build this select, maybe with a link to the issues and a short description of the problem as JS/TS-Doc.

timidak (changes requested):
Just a few general things, could you please test it again on every browser, after you implemented the changes.

timidak (approved):
LGTM 👍🏻

PR #362, opened 2023-09-12
fix: Make Registration more usable on mobile

Jonasdoubleyou:
On mobile we make the header much smaller, giving more space for actual content


Fabian5150 (approved):
Looks good to me :)

PR #373, opened 2023-09-28
feat: show email of pupils in screening dashboard

Leo310:
Resolves https://github.com/corona-school/project-user/issues/925

### DIFF
--- a/src/pages/screening/Dashboard.tsx
+++ b/src/pages/screening/Dashboard.tsx
@@ -26,6 +26,9 @@ function PupilCard({ pupil, onClick }: { pupil: PupilForScreening; onClick: () =
                             {pupil.firstname} {pupil.lastname}
                         </Heading>
                         <Text color="white" fontSize="15px">
+                            E-Mail: <b>{pupil.email}</b>
+                        </Text>
+                        <Text color="white" fontSize="15px">
                             {t('screening.registered_since')} {new Date(pupil!.createdAt).toLocaleDateString()}
                         </Text>
                     </VStack>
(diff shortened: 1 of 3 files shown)


realmayus (approved):

src/pages/screening/Dashboard.tsx
```diff
@@ -207,10 +212,12 @@ export function ScreeningDashboard() {
                         {disputedScreenings && disputedScreenings.pupilsToBeScreened.length !== 0 && (
                             <>
                                 <Heading>{t('screening.disputed_screenings')}</Heading>
-                                <HStack marginTop="20px">
+                                <VStack marginTop="20px">
```
Jonasdoubleyou, line 215: I think this should rather be a HStack with flex & flex-wrap. See my change ;)

Jonasdoubleyou (approved):

src/pages/screening/Dashboard.tsx
Leo310: Where can I see your change? Because with HStack it just overflows on the x-axis.
But I agree, with flex-wrap that would be fixed.

Jonasdoubleyou:
you'll now see it in the conflicts ;)

Jonasdoubleyou (approved):

src/pages/screening/Dashboard.tsx
Jonasdoubleyou: Ah, I think I forgot the disputed screenings and only fixed the search result UI :/ Anyways, HStack should work for now

PR #374, opened 2023-09-28
Add homeworkhelp after summerbreak.

Maascha:
Resolves https://github.com/corona-school/project-user/issues/944


src/pages/pupil/Dashboard.tsx
```diff
@@ -373,6 +375,20 @@ const Dashboard: React.FC<Props> = () => {
                                     ))) || <AlertMessage content={t('dashboard.noproposalsPupil')} />}
                             </HSection>
                         </VStack>
+                        <VStack marginBottom={space['1.5']}>
```
Jonasdoubleyou, line 378: Maybe we can hide the whole section if the env var is empty? That way we have a toggle in the future, and we avoid the scenario where the env is missing and users click on a non-working button.

Jonasdoubleyou (dismissed):

Jonasdoubleyou (approved):

PR #375, opened 2023-09-30
Fix argument oder of toTimerString invocation, improve days>7 string

realmayus:
Fix argument oder of toTimerString invocation, improve days>7 string to show actual date if time is in the week after next week.


src/Utility.ts
```diff
@@ -11,7 +11,8 @@ export const REDIRECT_PASSWORD = `/login`;
 
 export const toTimerString = (refDate: DateTime, compareDate: DateTime) => {
     const days = compareDate.startOf('day').diff(refDate.startOf('day'), 'days').days;
-    if (days > 7) return `In einigen Wochen`;
+    if (days > 7 && days < 14) return `Nächste Woche`;
```
Jonasdoubleyou, line 14: Could also be Übernächste Woche

Jonasdoubleyou (approved):

PR #379, opened 2023-10-14
Add user impersonation button if screener has TRUSTED_SCREENER role

realmayus:
Add user impersonation button if screener has TRUSTED_SCREENER role
Frontend part of https://github.com/corona-school/project-user/issues/928

I had to implement a workaround for the "open in new tab" functionality as it was blocked by the browser's popup blocker (apparently you can only call it in synchronous onClick event handlers..)

### DIFF
--- a/src/widgets/screening/ScreenPupilCard.tsx
+++ b/src/widgets/screening/ScreenPupilCard.tsx
@@ -236,6 +236,9 @@ export function ScreenPupilCard({ pupil, refresh }: { pupil: PupilForScreening; 
     // Otherwise the screening is successful and not invalidated yet, so no need to take action
 
     const impersonate = async () => {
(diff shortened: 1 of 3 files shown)


src/widgets/screening/ScreenPupilCard.tsx
```diff
@@ -228,6 +235,20 @@ export function ScreenPupilCard({ pupil, refresh }: { pupil: PupilForScreening;
 
     // Otherwise the screening is successful and not invalidated yet, so no need to take action
 
+    const impersonate = async () => {
+        const w = window.open('', '_blank');
```
Jonasdoubleyou, line 239: I think a comment would be good to explain why

Jonasdoubleyou (dismissed):

PR #382, opened 2023-10-15
753 daz matching

Fabian5150:
closes https://github.com/corona-school/project-user/issues/753


src/widgets/SubjectSelector.tsx
```diff
@@ -24,7 +24,7 @@ export const SubjectSelector = ({
 
     return (
         <HStack w="100%" flexWrap="wrap" justifyContent="center" alignItems="center">
-            {((selectable ?? SUBJECTS.filter((it) => it !== DAZ)) as string[]).map((subject) => (
+            {((selectable ?? SUBJECTS) as string[]).map((subject) => (
```
Jonasdoubleyou, line 27: Is this component used somewhere else, were we still want to exclude DaZ (pupil match request)?

src/types/subject.ts
```diff
@@ -32,7 +32,7 @@ export const SUBJECTS = [
 
 // Unfortunately this mapping is necessary as paths need to be ASCII to work reliably in the build setup
 export const SUBJECT_TO_ICON: { [subject in typeof SUBJECTS[number]]: string } = {
-    'Deutsch als Zweitsprache': 'deutsch_als_zweitsprace',
+    'Deutsch als Zweitsprache': 'deutsch_als_zweitsprache',
```
Jonasdoubleyou, line 35: Was this wrong before? Is the file named correctly?

Fabian5150: Yes, the file is named correctly, it was just misspelled here. Before that it didn't map the icon svg, now it does.
Was probably unnoticed before, as the icon wasn't used until now.

src/widgets/SubjectSelector.tsx
Fabian5150: Oh... good call. 
I checked all refrences of the component and in some it made sense, in others it didn't.
I would just create a new optional prop "includeDaz" or something like that.

Jonasdoubleyou (approved):
nice!

PR #383, opened 2023-10-15
732 add onboarding buttons to settings and help page

Fabian5150:
Closes https://github.com/corona-school/project-user/issues/732

Used a button in the help tab instead of a new tab, as the tabs require an iframe and there was already a outcommented button in the Helpcenter component:

grafik


Jonasdoubleyou (approved):
Please only merge if the onboarding contains the new content

PR #384, opened 2023-10-15
feat(dissolve-match-modal): showing warning modal for first 14 days

Leo310:
Resolves https://github.com/corona-school/project-user/issues/948

Further Comments/Questions:
1. the DissolvedMatchModal is also referenced but not used in the Student- and Pupil Dashboard. Should I remove them there or will they be used in the future?
2. The currentIndex is not set inside the Modal:
image

so that part is unreachable:
image
and could be removed.

### DIFF
--- a/src/lang/matching/de.ts
+++ b/src/lang/matching/de.ts
@@ -163,6 +163,11 @@ const matching = {
             title: 'Warum möchtest du das Lernpaar auflösen?',
             btn: 'Lernpaar auflösen',
         },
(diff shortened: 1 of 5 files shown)


Jonasdoubleyou:
(1) & (2) remove them. Let's readd it when we actually need it

Jonasdoubleyou (dismissed):
nice!

Jonasdoubleyou (approved):

PR #385, opened 2023-10-15
make tags mandatory when creating a course (#950)

paul-jaworski:


Jonasdoubleyou (commented):
Um, what are you trying to accomplish with this change?

src/types/lernfair/Course.ts
```diff
@@ -10,7 +10,7 @@ export type LFCourse = {
     id?: string;
     name: string;
     description: string;
-    tags?: LFTag[];
+    tags: LFTag[];
```
Jonasdoubleyou, line 13: I'm not sure that in all the places where the LFCourse type is still used we properly query all fields, so I would prefer to not change this type if we're not certain that this is not null in all cases.

paul-jaworski:
made tags optional again :)

Jonasdoubleyou (approved):

PR #387, opened 2023-10-15
fix: Always show user impersonation button

Jonasdoubleyou:
A trusted screener might want to log in with the user's account regardless of it's current state.

Also hide it for deactivated accounts, as login won't work then anyways.


Fabian5150 (approved):
Looks good to me

PR #392, opened 2023-10-28
Use new enum dissolve reasons

realmayus:
Frontend part of https://github.com/corona-school/project-user/issues/494


src/modals/DissolveMatchModal.tsx
```diff
@@ -1,30 +1,37 @@
...
+        Dissolve_Reason.NoMoreTime,
+        Dissolve_Reason.PersonalIssues,
+        Dissolve_Reason.ScheduleIssues,
+        Dissolve_Reason.TechnicalIssues,
+        Dissolve_Reason.LanguageIssues,
+        Dissolve_Reason.Other,
+    ];
 
+    const onReasonChange = (key: String) => {
+        setReason(Object.values(Dissolve_Reason).find((x) => x === key));
```
Jonasdoubleyou, line 33: Why not setReason(x) ?

Jonasdoubleyou (approved):

src/modals/DissolveMatchModal.tsx
realmayus: I added this because the `reason` state is of type `Dissolve_Reason`, but the radio button only stores its value as a string, so we somehow need to convert from a string to the enum

Jonasdoubleyou: key as DissolveReason

Jonasdoubleyou (approved):

PR #395, opened 2023-11-02
959 alternatives after matchrequest

Fabian5150:
Closes https://github.com/corona-school/project-user/issues/959
Added some generalization (as there was some copy-paste code) and i18next strings for the files that have been edited.


src/pages/pupil/matching/Filter.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+                    <Text>{t('matching.wizard.pupil.filter.text1')}</Text>
+
+                    <Box my={space['0.5']}>
+                        <BulletList bulletPoints={t('matching.wizard.pupil.filter.bulletPoints', { returnObjects: true })} />
+                    </Box>
+
+                    <Heading fontSize="md">{t('matching.wizard.pupil.filter.question1')}</Heading>
+                    <YesNoSelector
+                        initialYes={isFit === 'yes'}
```
Jonasdoubleyou, line 43: Can't the initialYes / initialNo be omitted?

Jonasdoubleyou (approved):

src/pages/pupil/matching/Filter.tsx
Fabian5150: I've tried it out: Appartenlty not, if so it is possible to select both "yes" and "no"

PR #398, opened 2023-11-04
feat: Semiautomatic Translations

Jonasdoubleyou:
Moves all existing translations into one single JSON file, to simplify automatic handling of translation files.

Adds a small script `npm run translate` which diffs the de.json file with other translation files, and automatically translates it using Weglot (or maybe also DeepL or Google Translation API?).

The script can also be run in Check Mode `npm run translate -- check` where it just diffs and fails if there are differences, this is run inside of a Github Action to ensure that all translations match.


Fabian5150 (approved):
A bit of a shame that we loose the individual translation files, but I think it's a good trade-off for simplifying the translation step. :)

Jonasdoubleyou:
@Maascha @realmayus what do you think about this change?

Maascha:
Does  

>  Existing entries in the translation files should not be changed when changing the UI, instead add new entries to the language file and delete the old entries (for the automatic translation to diff properly).

mean, that I cannot update anymore in the de.ts file, because its translation are not updated? That might lead to unintentional bugs. Is there some workaround that it still updates the translations then (something like a full check?). Weglot was kinda buggy previously on gendered language (Schüler:innen -> pupil:inside) and we could manually specify those. Is that still possible via the ui?

No objection against storing everything in one file.

Jonasdoubleyou:
We could add something like `npm run translate <key>` to translate a certain entry again, however it won't be picked up automatically by the script. We could also detect it through the git history, but that would complicate the script.

And yes, manually maintained translations in Weglot shall be respected.

Jonasdoubleyou:
merged this to avoid conflicts with the translation file, we can still discuss details of the translate script

realmayus:
Wow, very nice! 🎉
Perhaps we could've kept the files separate and construct the total json in memory, but I don't think it matters much. We could add a check ensuring any template patterns are still in the translated string, as these AI-based translators sometimes forget them/mess them up.
If they end up translating e.g. "Nachricht an {{pupil}} gesendet" to "Повідомлення надіслано {{учень}}" (deepL does this, at least), we could replace them with "X", "Y", ... but maybe there's an option in weglot to ignore such patterns if we use that.

PR #406, opened 2023-11-10
Fix language selection

realmayus:
- don't allow users to select french/turkish multiple times
- re-add turkish as a language (git commit history implies this was only excluded as a "bugfix")

### DIFF
--- a/src/lang/de.json
+++ b/src/lang/de.json
@@ -195,6 +195,8 @@
             "englisch": "Englisch",
             "franz_sisch": "Französisch",
             "französisch": "Französisch",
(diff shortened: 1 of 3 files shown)


src/lang/de.json
```diff
@@ -195,6 +195,8 @@
             "englisch": "Englisch",
             "franz_sisch": "Französisch",
             "französisch": "Französisch",
+            "türkisch": "Türkisch",
+            "t_rkisch": "Türkisch",
```
Jonasdoubleyou, line 199: why twice?

Jonasdoubleyou (approved):

PR #410, opened 2023-11-13
feat: Export / Import Translations & Template Handling

Jonasdoubleyou:
Escape templates ({{something}} -> {{0}}) prior to sending to the Translation Provider, so that the template variable is not accidentally picked up by the translator. Might not be ideal for languages with a different sentence structure than german, but for english it seems to work reasonably well.

Also provides `npm run translate -- (export|import)` with which one can import a TXT file with all changed translations, and reimport from that text file. The text file is meant to be sent to human translators and is generally "editable" (correct translations can be deleted, wrong translations can just be edited).


realmayus (approved):
Nice!

PR #415, opened 2023-11-26
feat: Scroll Course Cards back into View

Jonasdoubleyou:


Fabian5150:
There is currently a bug, where if there are multiple courses next to each other (horizontally) in the list, only the first one will be rendered again, after going to one course and than using the back button to get back to /group.
It's also possible that this bug already existed before, but wasn't noticed yet. Anyway I will look into it :eyes:

Fabian5150:
The problem with the second course card not being rendered seemed to only happen with course 3, I could not reproduce it with any other course. (I'm still trying to figure out why)
But now I've seen another bug, where when navigating: "my courses" tab -> one of the courses there -> back to "my courses" suddenly a course gets rendered, where the logged in pupil isn't a participant. When refreshing the page it disappears again.
It's getting stranger with every minute.

Fabian5150:
The mentioned bugs appear to have "accidentally" been fixed along the way with https://github.com/corona-school/user-app/pull/440

The former local course search was faulty and used to cause this strange behavior.

Fabian5150 (approved):

PR #421, opened 2023-12-13
feat: add join match meeting mutation

LomyW:
# Description
We need to add the mutation that triggers a join match meeting event for achievement at any point in the user app where the user clicks on the VideoButton. To do this, we need the matchId (or later subcourseId) from the appointment.

## ✅ What was done
- add mutation to the `VideoButton` to trigger the join match meeting event
- adjust appointment queries 


LucasFalkowsky (approved):
simple enough! Well done!

PR #423, opened 2023-12-15
Feat: next step cards

LucasFalkowsky:
What was done:

- addition of nextStepAchievements to ImportantInformation query
- change to NextStepCards
- addition of different modals for important information and next step achievements

still missing:

- icon discriminator for different important information types


LomyW (changes requested):
some design changes 🎨

src/helper/achievement-helper.ts
```diff
@@ -14,14 +22,26 @@ function checkAndGetSecondEnumValue<T extends Record<string, string>>(enumElemen
...
-    data: InactiveAchievementsQuery;
+    data?: InactiveAchievementsQuery;
 };
-function convertDataToAchievement(dataWithType: AchievementsQueryType | InactiveAchievementsQueryType): Achievement[] {
-    const query = dataWithType.type === 'achievements' ? dataWithType.data.me.achievements : dataWithType.data.me.inactiveAchievements;
+export type NextStepAchievementsQueryType = {
+    type: 'nextStepAchievements';
+    data?: GetOnboardingInfosQuery;
+};
+function convertDataToAchievement(dataWithType: AchievementsQueryType | InactiveAchievementsQueryType | NextStepAchievementsQueryType): Achievement[] {
```
LomyW, line 35: maybe rename function to `convertAchievementToNextStep` and dataWithType to `achievementWithType`? Would specify it more to the usecase.

src/helper/achievement-helper.ts
```diff
@@ -14,14 +22,26 @@ function checkAndGetSecondEnumValue<T extends Record<string, string>>(enumElemen
 
 export type AchievementsQueryType = {
     type: 'achievements';
```
LomyW, line 24: should type maybe a enum?

src/helper/achievement-helper.ts
LomyW: Is this something we cant handle in the resolver?

src/widgets/ImportantInformation.tsx
```diff
@@ -344,75 +377,101 @@ const ImportantInformation: React.FC<Props> = ({ variant }) => {
...
+                        <NextStepsCard
+                            key={index}
+                            title={info.title}
+                            name={`${t('important')}!`}
+                            description={info.desciption}
+                            actionDescription={t('moreInfoButton')}
+                            actionType={Achievement_Action_Type_Enum.Action}
+                            onClick={() => {
+                                info.btnfn && info.btnfn();
+                                console.log('CLICKED!');
```
LomyW, line 437: can you please remove the log here?

src/widgets/ImportantInformation.tsx
```diff
@@ -344,75 +377,101 @@ const ImportantInformation: React.FC<Props> = ({ variant }) => {
...
+                scrollable
+                title={t('helperwizard.nextStep')}
+                referenceTitle={t('helperwizard.progress')}
+                marginBottom="25px"
+                onShowAll={() => navigate('/progress')}
+                showAll
+            >
+                {configurableInfos.map((info, index) => {
+                    return (
+                        <NextStepsCard
```
LomyW, line 428: for now we do not pass the `actionRedirectLink`` or?

src/widgets/ImportantInformation.tsx
```diff
@@ -344,75 +377,101 @@ const ImportantInformation: React.FC<Props> = ({ variant }) => {
...
+                            key={index}
+                            title={info.title}
+                            name={`${t('important')}!`}
+                            description={info.desciption}
+                            actionDescription={t('moreInfoButton')}
```
LomyW, line 433: Do we have a more precise description here?

src/widgets/ImportantInformation.tsx
```diff
@@ -344,75 +377,101 @@ const ImportantInformation: React.FC<Props> = ({ variant }) => {
...
+                                info.btnfn && info.btnfn();
+                                console.log('CLICKED!');
+                            }}
+                        />
+                    );
+                })}
+                {infos.map((config, index) => {
+                    const buttontexts: string[] = t(`helperwizard.${config.label}.buttons` as unknown as TemplateStringsArray, { returnObjects: true });
+                    return (
+                        <NextStepsCard
```
LomyW, line 445: also here we dont pass the `actionRedirectLink`

src/components/achievements/nextStepsCard/NextStepsCard.tsx
```diff
@@ -36,39 +51,49 @@ const NextStepsCard: React.FC<NextStepsCardProps> = ({ image, title, name, actio
...
-                )}
-            </HStack>
-            <HStack alignItems="flex-start" space="4px" justifyContent="flex-start">
-                {actionType && (
-                    <VStack height="19px" position="relative" justifyContent="center">
-                        <Box width="8px" height="8px">
-                            {icon}
-                        </Box>
+        <Pressable onPress={onClick} disabled={!onClick}>
+            <Link width="288px" height="288px" href={actionRedirectLink}>
```
LomyW, line 55: I think the link on the card itself should not be clickable, only the whole card.

src/components/achievements/nextStepsCard/NextStepsCard.tsx
```diff
@@ -36,39 +51,49 @@ const NextStepsCard: React.FC<NextStepsCardProps> = ({ image, title, name, actio
...
+                <VStack width="100%" height="100%" padding="24px" backgroundColor="primary.900" borderRadius="8px" justifyContent="space-between">
+                    <VStack width="fit-content" borderRadius="8px">
+                        {image ? (
+                            <Image width="64px" height="64px" src={image} />
+                        ) : (
+                            <Box width="64px" height="64px">
+                                <BooksIcon />
```
LomyW, line 62: We should think about different Icons for the `ImportantInformations`

LomyW (changes requested):
some changes 😇

src/widgets/ImportantInformation.tsx
```diff
@@ -344,75 +377,101 @@ const ImportantInformation: React.FC<Props> = ({ variant }) => {
...
+                marginBottom="25px"
+                onShowAll={() => navigate('/progress')}
+                showAll
+            >
+                {configurableInfos.map((info, index) => {
+                    return (
+                        <NextStepsCard
+                            key={index}
+                            title={info.title}
+                            name={`${t('important')}!`}
```
LomyW, line 431: I think this is `info.title`. I guess the actual `title` (what is the subtitle) depends on the information type - which does not yet exist. But I think the type would make sense to a) get the icons and b) get the subtitles.

LomyW (changes requested):
further idea 😁

src/helper/achievement-helper.ts
```diff
@@ -1,4 +1,12 @@
...
+    Achievement_State,
+    Achievement_Type_Enum,
+    AchievementsQuery,
+    GetOnboardingInfosQuery,
+    InactiveAchievementsQuery,
+    Step,
+} from '../gql/graphql';
 import { Achievement } from '../types/achievement';
 
 function checkAndGetSecondEnumValue<T extends Record<string, string>>(enumElement: any, comparator: T): keyof T | null {
```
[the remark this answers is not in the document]
LucasFalkowsky: This used to make sense when we had different enums in front and backend. But you are right. This function should be deleted by now. It was removed.

src/helper/achievement-helper.ts
LucasFalkowsky: This function is used on progress page and as next step so renaming would not make sense here. I renamed the incoming data. No this is resolving a typing issue that we have, when getting the data from the backend.

src/helper/achievement-helper.ts
LucasFalkowsky: Good objection!

src/widgets/ImportantInformation.tsx
LucasFalkowsky: Yes you are right I forgot to include it in the modal. the redirect link is used by the modal now, but still not the card, because that would be conflicting with the card being a modal opener.

src/widgets/ImportantInformation.tsx
LucasFalkowsky: The more precise description would require an extra field in the database, if the usage is supposed to be easy. The button descriptions are used in the modal it self.
I added the possibility to write a description in the i18n file and if not, it takes moreInfoButton.

src/widgets/ImportantInformation.tsx
LucasFalkowsky: for the same reason. The redirect for this would be handled by the whole shabang above the return (the btnfn would redirect you in this case).

src/components/achievements/nextStepsCard/NextStepsCard.tsx
LucasFalkowsky: Thats what is happening here. We had a Pressable and if the onClick is undefined we would have used the Link.
I removed the Link anyway since the redirect would be handled by the parent component so this is not necessary here.

LomyW (commented):
questions 😇

src/helper/achievement-helper.ts
LucasFalkowsky: the type graphql returns is not the type we pass in the resolver

src/components/achievements/modals/NextStepModal.tsx
```diff
@@ -0,0 +1,60 @@
...
+                            <Text fontSize="36px" bold>
+                                {title}
+                            </Text>
+                        </VStack>
+                    </HStack>
+                    <Box>
+                        <Text>{description}</Text>
+                    </Box>
+                    <Box>
+                        <Stack space={4} width="100%" direction={buttons?.length && buttons.length > 1 ? 'row' : 'column'} flexWrap="wrap">
```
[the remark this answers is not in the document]
LucasFalkowsky: the buttons being in a row works as long as there are just 2 buttons. 3 buttons would make the modal overflow in a lot of cases. But since we are talking about .length and not index, the number should be > 2 not >1 so this was fixed.

src/helper/achievement-helper.ts
```diff
@@ -1,32 +1,35 @@
...
-    if (typeof enumElement !== 'string') {
-        return null;
-    }
-    const secondEnumElement = Object.keys(comparator).find((key) => comparator[key] === enumElement);
-    if (secondEnumElement) {
-        return secondEnumElement;
-    }
-    return null;
+export enum TypeofAchievementQuery {
+    achievements = 'achievements',
```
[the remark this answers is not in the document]
LucasFalkowsky: I originaly wanted to do that but when I tried to read FURTHERACHIEVEMENTS and NEXTSTEPACHIEVEMENTS I gave up on that idea. is this ok of a reason or you think I should change it?

src/widgets/ImportantInformation.tsx
```diff
@@ -1,23 +1,34 @@
...
+import { Achievement } from '../types/achievement';
+import AchievementModal from '../components/achievements/modals/AchievementModal';
+import { TypeofAchievementQuery, convertDataToAchievement } from '../helper/achievement-helper';
+import NextStepModal from '../components/achievements/modals/NextStepModal';
 
 type Props = {
     variant?: 'normal' | 'dark';
 };
 
+type Information = {
```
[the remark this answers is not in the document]
LucasFalkowsky: should we add it already or should we wait for a ticket that makes this necessary of a change.

src/widgets/ImportantInformation.tsx
```diff
@@ -344,75 +377,104 @@ const ImportantInformation: React.FC<Props> = ({ variant }) => {
...
+                            onClick={() => {
+                                info.btnfn && info.btnfn();
+                            }}
+                        />
+                    );
+                })}
+                {infos.map((config, index) => {
+                    const buttontexts: string[] = t(`helperwizard.${config.label}.buttons` as unknown as TemplateStringsArray, { returnObjects: true });
+                    const actionDescription =
+                        t(`helperwizard.${config.label}.actionDescription` as unknown as TemplateStringsArray, config.lang) || t('moreInfoButton');
```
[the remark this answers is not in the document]
LucasFalkowsky: nope because they did not exist prior to this update.

src/widgets/ImportantInformation.tsx
LucasFalkowsky: I found that there is a i18n.exists function is used for such cases and implemented it with that.

LomyW (approved):
looks good to me 💯

PR #424, opened 2023-12-23
1004 tooltips disabled buttons

Fabian5150:
Closes https://github.com/corona-school/project-user/issues/1004

Lots of small stuff

Added generalized Component components/DisablebleButton to automatically render a tooltip around a button if disabled.
This component requires a "reasonDisabled" field so all Buttons, that can be disabled (assuming they use this new component) have provide one.
This way the other components stay a bit cleaner and more unified
Than changed all accuring Buttons, that can be disabled to use this component.
Other button props now have to be assigned slightly different, all as one object given to DisablebleButton, but it does not sacrifice any functionality or creates significantly more effort/complexitiy.


src/components/DisablebleButton.tsx
```diff
@@ -0,0 +1,23 @@
...
+type ButtonProps = ComponentProps<typeof Button>;
+
+type DisablebleButtonProps = {
+    isDisabled: boolean;
+    reasonDisabled: string;
+    buttonProps?: ButtonProps;
+    disabledButtonContent?: ReactNode;
+};
+
+const DisablebleButton: React.FC<PropsWithChildren<DisablebleButtonProps>> = ({ isDisabled, reasonDisabled, disabledButtonContent, children, buttonProps }) => {
```
Jonasdoubleyou, line 13: DisableableButton?

src/components/DisablebleButton.tsx
```diff
@@ -0,0 +1,23 @@
+import { Button, Tooltip } from 'native-base';
+import { PropsWithChildren, ComponentProps, ReactNode } from 'react';
+
+type ButtonProps = ComponentProps<typeof Button>;
+
+type DisablebleButtonProps = {
```
Jonasdoubleyou, line 6: = { ... } & ButtonProps; and then pass them on as spread properties (...rest in parameters, {...rest} in component)? Or any reason to have "buttonProps" on a component named ...Button?

Jonasdoubleyou (commented):
nice refactoring, some minor improvements to be made to the button component

src/components/DisablebleButton.tsx
Fabian5150: I don't really like the name either... But I couldn't come up with something better, I think just "DisabledButton" or something like that would give a wrong impression about what the component is actually there for

src/components/DisablebleButton.tsx
Fabian5150: Should've known that earlier... Gave up to earlier on finding a better way to pass the props down to the button, but it's in the new commit now. :)

src/components/DisablebleButton.tsx
Fabian5150: Oh... will fix that real quick here

Jonasdoubleyou (approved):

PR #430, opened 2024-01-04
Make tag selection mandatory

realmayus:
Closes https://github.com/corona-school/project-user/issues/950


realmayus:
Just realized this is already gets implemented in https://github.com/corona-school/user-app/pull/385 😅

PR #437, opened 2024-01-14
Skip subject prio page if only one subject selected

Fabian5150:
Closes https://github.com/corona-school/project-user/issues/1080

### DIFF
--- a/src/pages/pupil/matching/Subjects.tsx
+++ b/src/pages/pupil/matching/Subjects.tsx
@@ -15,9 +15,10 @@ const Subjects: React.FC = () => {
     const isDAZ = containsDAZ(matchRequest.subjects);
 
     useEffect(() => {
-        setSkippedSubjectPriority(isDAZ || matchRequest.subjects.length === 1);
-        matchRequest.subjects.forEach((subj) => setSubjectPriority(subj.name, skippedSubjectPriority));
-    }, [matchRequest.subjects.length, isDAZ, setSkippedSubjectPriority, setSubjectPriority, skippedSubjectPriority]);
+        const skipSubjectPriority = isDAZ || matchRequest.subjects.length === 1;
+        setSkippedSubjectPriority(skipSubjectPriority);
+        matchRequest.subjects.forEach((subj) => setSubjectPriority(subj.name, skipSubjectPriority));
+    }, [matchRequest.subjects.length, isDAZ, setSkippedSubjectPriority, setSubjectPriority]);
 
     return (
         <VStack paddingX={space['1']} space={space['0.5']}>
--- a/src/pages/pupil/matching/Subjects.tsx
+++ b/src/pages/pupil/matching/Subjects.tsx
@@ -16,11 +16,8 @@ const Subjects: React.FC = () => {
 
     useEffect(() => {
         setSkippedSubjectPriority(isDAZ || matchRequest.subjects.length === 1);
-    }, [matchRequest.subjects.length, isDAZ]);
-
-    useEffect(() => {
         matchRequest.subjects.forEach((subj) => setSubjectPriority(subj.name, skippedSubjectPriority));
-    }, [setSubjectPriority, skippedSubjectPriority]);
+    }, [matchRequest.subjects.length, isDAZ, setSkippedSubjectPriority, setSubjectPriority, skippedSubjectPriority]);
 
     return (
         <VStack paddingX={space['1']} space={space['0.5']}>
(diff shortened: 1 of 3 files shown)


src/pages/pupil/matching/Subjects.tsx
```diff
@@ -8,19 +8,28 @@ import { useTranslation } from 'react-i18next';
...
-    const { matchRequest, setSubject, removeSubject, setCurrentIndex } = useContext(RequestMatchContext);
+    const { matchRequest, setSubject, removeSubject, setCurrentIndex, setSkippedSubjectPriority, skippedSubjectPriority, setSubjectPriority } =
+        useContext(RequestMatchContext);
     const { t } = useTranslation();
 
     const isDAZ = containsDAZ(matchRequest.subjects);
 
+    useEffect(() => {
+        setSkippedSubjectPriority(isDAZ || matchRequest.subjects.length === 1);
+    }, [matchRequest.subjects.length]);
```
Jonasdoubleyou, line 19: isDaz missing in dependencies

src/pages/pupil/matching/Subjects.tsx
```diff
@@ -8,19 +8,28 @@ import { useTranslation } from 'react-i18next';
...
+
+    useEffect(() => {
+        matchRequest.subjects.forEach((subj) => setSubjectPriority(subj.name, skippedSubjectPriority));
+    }, [setSubjectPriority, skippedSubjectPriority]);
```
Jonasdoubleyou, line 23: Do we need these cascading effects? Can't this be one effect?

src/pages/pupil/matching/Subjects.tsx
Fabian5150: I didn't add it because its value doesn't get changed while this component is rendered.
But you're right, I should add it. It might get changed in different places in the future...

src/pages/pupil/matching/Subjects.tsx
Fabian5150: It can. I've updated it and it works without a problem or running the functions too often.
:) :+1:

src/pages/pupil/matching/Subjects.tsx
```diff
@@ -16,11 +16,8 @@ const Subjects: React.FC = () => {
 
     useEffect(() => {
         setSkippedSubjectPriority(isDAZ || matchRequest.subjects.length === 1);
-    }, [matchRequest.subjects.length, isDAZ]);
-
-    useEffect(() => {
         matchRequest.subjects.forEach((subj) => setSubjectPriority(subj.name, skippedSubjectPriority));
-    }, [setSubjectPriority, skippedSubjectPriority]);
+    }, [matchRequest.subjects.length, isDAZ, setSkippedSubjectPriority, setSubjectPriority, skippedSubjectPriority]);
```
Fabian5150, line 20: It's btw. necesarry/better that it's matchRequest.subjects.length here and not just matchRequest.subjects, as this would cause much much more rerenders I found out.
The functinality should not be affected by that

src/pages/pupil/matching/Subjects.tsx
```diff
@@ -8,19 +8,25 @@ import { useTranslation } from 'react-i18next';
...
-    const { matchRequest, setSubject, removeSubject, setCurrentIndex } = useContext(RequestMatchContext);
+    const { matchRequest, setSubject, removeSubject, setCurrentIndex, setSkippedSubjectPriority, skippedSubjectPriority, setSubjectPriority } =
+        useContext(RequestMatchContext);
     const { t } = useTranslation();
 
     const isDAZ = containsDAZ(matchRequest.subjects);
 
+    useEffect(() => {
+        setSkippedSubjectPriority(isDAZ || matchRequest.subjects.length === 1);
+        matchRequest.subjects.forEach((subj) => setSubjectPriority(subj.name, skippedSubjectPriority));
```
Jonasdoubleyou, line 19: skippedSubjectPriority refers to the previous value. I'd prefer:

```
const skipSubjectPriority = isDAZ || matchRequest.subjects.length === 1;
setSkippedSubjectPriority(skipSubjectPriority);
... setSubjectPriority(subj.name, skipSubjectPriority)
```

Also I wonder if this works currently?

Fabian5150: It did work, but I've changed it to your suggestion now, that does look better... :)

Jonasdoubleyou (approved):
CI failure can be ignored, see my fix ...

PR #442, opened 2024-01-15
Feat: achievements empty states

LucasFalkowsky:
closes https://github.com/corona-school/project-user/issues/1067

### What was done:

- empty state component was introduced


src/components/achievements/EmptyStateContainer.tsx
```diff
@@ -0,0 +1,30 @@
...
+};
+
+const EmptyStateContainer = ({ achievementState }: EmptyStateContainerProps) => {
+    const { t } = useTranslation();
+    switch (achievementState) {
+        case Achievement_State.Completed:
+            return (
+                <VStack width="100%" display="flex" flexDirection="column" alignItems="center" space="4px">
+                    <EmptyStateCompleted />
+                    <Text fontWeight={600} fontSize="14px" lineHeight="17px" textAlign="center">
```
LomyW, line 17: can we please use the pre-defined fontSizes like 'xs', 'sm' etc.?

src/components/achievements/EmptyStateContainer.tsx
```diff
@@ -0,0 +1,30 @@
...
+                        {t('achievement.empty_state.Completed.title')}
+                    </Text>
+                    <Text fontWeight={400} fontSize="10px" lineHeight="12px" textAlign="center">
```
LomyW, line 20: also here? I think 10px is little bit too small

LomyW, line 20: also here, think 10px is little bit to small

src/components/achievements/EmptyStateContainer.tsx
```diff
@@ -0,0 +1,30 @@
...
+                        {t('achievement.empty_state.Completed.description')}
+                    </Text>
```
LomyW, line 22: maybe something like that? (that it looks similar to the other empty states in the user app i.e. `AppointmentEmptyState``)
```suggestion
                     <Heading>{t('achievement.empty_state.Completed.title')}</Heading>
                    <Text maxW="700" textAlign="center" py="3">
                        {t('achievement.empty_state.Completed.description')}
                    </Text>
```

src/widgets/AchievementProgress.tsx
```diff
@@ -191,46 +192,58 @@ const AchievementProgress: React.FC<AchievementProgressProps> = ({ achievements,
                 {states.map((key) => (
                     <VStack key={key} space={3} marginTop={10}>
                         <ProgressCollapsableHeadline achievementState={key} onClick={() => handleOnClick(undefined, key)} />
```
LomyW, line 194: is the empty state also collapsable?

LomyW (changes requested):
some notes ✨

src/components/achievements/EmptyStateContainer.tsx
LucasFalkowsky: It is exactly what the design sais. Thats why I used these values.
I changed them to the presets though.

src/components/achievements/EmptyStateContainer.tsx
LucasFalkowsky: But this is not as it was done in the design.

LomyW (approved):
Before merging please resolve the conflict.
Looks good to me. 🎉

PR #444, opened 2024-01-18
fix: adapt join meeting button time for organizers

LomyW:
# Description
The video chat button in the "Next appointment" section was not adjusted when switching to 4 hours
- Fixes https://github.com/corona-school/project-user/issues/1090

## ✅ What was done
- [x] increase time to 4h
 
- [x] Tested on different browsers.


Jonasdoubleyou:
That we have multiple places with the same logic suggests it should be refactored

PR #452, opened 2024-01-23
fix: hide empty chats

LomyW:
# Description
- Closes https://github.com/corona-school/project-user/issues/1091

We want to hide empty chat. Empty chats are chats where no "real" message where sent.
works with: https://github.com/corona-school/backend/pull/963
## ✅ What was done
- [x] set state for selected chat
 
- [x] Tested on different browsers.


src/pages/Chat.tsx
```diff
@@ -87,21 +86,28 @@ const Chat: React.FC = () => {
...
         if (!session) return;
+
         const inbox = session.createInbox({
             showMobileBackButton: false,
             messageField: { visible: { access: ['==', 'ReadWrite'] }, placeholder: t('chat.placeholder') },
         });
+
         inbox.mount(inboxRef.current);
-        inbox.select(conversationId ?? selectedChatId);
+        inbox.select(selectedChatId || conversationId);
```
dhenkel92, line 96: Did you use `||` on purpose?

LomyW: yes because conversationId can be undefined. The `conversationId` is only set if we navigate from the course or match page to the chat.

dhenkel92: The question was if you are using `||` instead of `??` on purpose.

LomyW: yes :)

Jonasdoubleyou (approved):

PR #489, opened 2024-03-04
Fix/coc button behaviour

Fabian5150:
Closes https://github.com/orgs/corona-school/projects/8/views/7?pane=issue&itemId=54524588


src/pages/CertificateOfConduct.tsx
```diff
@@ -53,8 +57,10 @@ const CertificateOfConduct = () => {
     const [downloadRemissionRequest] = useMutation(gql(`mutation DownloadRemissionRequest { studentGetRemissionRequestAsPDF }`));
 
     async function openRemissionRequest() {
```
Fabian5150, line 59: Only disables the Button while downloadRemissionRequest is pending, not the download itself.
But from testing it a bit it seems fine, because the request is what takes a few seconds, not the download itself.
Checking if the browser has started the download is a bit more tricky and felt a bit overkill here. The UX feels normal I would say. Even tested with slow Deutsche Bahn - Wifi and it expects as anticipated.

Although if wished I could implement the check for whether the download itself has started

Nevertheless if you think it's n

src/pages/CertificateOfConduct.tsx
```diff
@@ -53,8 +57,10 @@ const CertificateOfConduct = () => {
...
+        setDissableDownload(true);
```
Fabian5150, line 60: Only disables the Button while downloadRemissionRequest is pending, not the download itself.
But from testing it a bit it seems fine, because the request is what takes a few seconds, not the download itself.
Checking if the browser has started the download is a bit more tricky and felt a bit overkill here. The UX feels normal I would say. Even tested with slow Deutsche Bahn - Wifi and it expects as anticipated.

Although if wished I could implement the check for whether the download itself has started

src/pages/CertificateOfConduct.tsx
```diff
@@ -53,8 +57,10 @@ const CertificateOfConduct = () => {
...
         const { data } = await downloadRemissionRequest();
         downloadFile('Antrag_auf_Befreiung_von_Kosten.pdf', BACKEND_URL + data!.studentGetRemissionRequestAsPDF);
+        setDissableDownload(false);
```
Fabian5150, line 63: Only disables the Button while downloadRemissionRequest is pending, not the download itself.
But from testing it a bit it seems fine, because the request is what takes a few seconds, not the download itself.
Checking if the browser has started the download is a bit more tricky and felt a bit overkill here. The UX feels normal I would say. Even tested with slow Deutsche Bahn - Wifi and it expects as anticipated.

Although if wished I could implement the check for whether the download itself has started

src/pages/CertificateOfConduct.tsx
```diff
@@ -31,6 +33,8 @@ const CertificateOfConduct = () => {
     const { space } = useTheme();
     const cocDate = data?.me?.student?.certificateOfConductDeactivationDate;
 
+    const [dissableDownload, setDissableDownload] = useState(false);
```
Jonasdoubleyou, line 36: typo: disabled

src/pages/CertificateOfConduct.tsx
Jonasdoubleyou: Should be good enough! Also thanks for testing under realistic conditions ;)

Jonasdoubleyou (approved):

src/pages/CertificateOfConduct.tsx
Fabian5150: oops

Jonasdoubleyou (approved):

PR #493, opened 2024-03-09
fix: always generate device token

dhenkel92:
When using a secret token to loign (E-Mail / verifiction / etc.) you should get a device token as well. Otheriwse, the user will be logged out automatically every day.

### DIFF
--- a/src/hooks/useApollo.tsx
+++ b/src/hooks/useApollo.tsx
@@ -46,8 +46,6 @@ export type LFApollo = {
     client: ApolloClient<NormalizedCacheObject>;
     // Invalidated the device token, destroys the session and goes back to the start page
     logout: () => Promise<void>;
-    // Call in case a mutation was run that associates the session with a user
-    onLogin: (result: FetchResult) => void;
     // If we suspect that a backend mutation caused a change in user data or roles, we might want to refresh:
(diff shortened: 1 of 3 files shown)


src/hooks/useApollo.tsx
```diff
@@ -347,14 +347,15 @@ const useApolloInternal = () => {
                 log('GraphQL', 'successfully logged in with secret token');
                 setSessionState('logged-in');
                 setUser(null); // refresh user information
+                createDeviceToken(); // fire and forget
```
Jonasdoubleyou, line 350: Can you quickly check where we call onLogin? That was supposed to create it

Jonasdoubleyou (approved):

src/hooks/useApollo.tsx
dhenkel92: AFAICS here

https://github.com/corona-school/user-app/blob/1c811c59f97e546eaccd24bd71a2b583988f107b/src/pages/Login.tsx#L168-L173

Jonasdoubleyou: Maybe we should remove onLogin entirely then and just put it into loginWithPassword

Jonasdoubleyou (approved):

PR #495, opened 2024-03-11
Make support offers more present

JeangelLF:
## What was done

- Created a SupportSection component as described in https://github.com/corona-school/project-user/issues/960
- Added needed translation strings
- Added needed icons _(I grabbed them directly from the landing page, not sure if we have a Figma file with those or something similar)_

**Here is how it looks:**
Bildschirmfoto 2024-03-11 um 18 05 36

**Notes:**
- I didn't use the same tab ids provided in the issue, as some of those weren't working _(I also grabbed them from the landing page 😬)_
- Initially tried to use `IconTagList` to build this but that component is quite strict in how it works/looks. It's probably worth adjusting it or create another one more re-usable/flexible, but definitely for a separate task.


src/pages/student/dashboard/SupportSection.tsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
```
JeangelLF, line 1: Just when sending the PR, I'm kind of noticing that this should live under `src/widgets` maybe? If so, probably without the Heading 🤔

realmayus: Yeah, I think widgets would be a better fit. The headline can stay in the file imo

Jonasdoubleyou (approved):

Jonasdoubleyou (approved):

PR #497, opened 2024-03-15
refactor: Cleanup storybook components

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1146

## What was done?

- Removed a bunch of unused components
- Fixed some stories that were throwing errors
- Removed some unneeded stories. For these, I tried to move the comments from the Story to the component file to not lose that documentation

I also wanted to upgrade Storybook to the latest version but I'll tackle that one in a separate task


src/widgets/ListItem.tsx
```diff
@@ -1,29 +1,22 @@
...
 
 type Props = {
     label: string;
-    value?: string;
     onPress?: () => any;
     isDisabled?: boolean;
 };
 
-const EditDataRow: React.FC<Props> = ({ label, value, onPress, isDisabled }) => {
+const EditDataRow: React.FC<Props> = ({ label, onPress, isDisabled }) => {
```
Jonasdoubleyou, line 9: const ListItem

src/components/CSSWrapper.tsx
```diff
@@ -6,6 +6,10 @@ type Props = {
     style?: Object;
 };
 
+/**
+ * This wrapper component is used to apply css in places where native it not applicable or hard to achieve.
+ * There is a *.native file that was a fallback for natives contexts
+ */
```
Jonasdoubleyou, line 12: (I guess sooner or later we should finally decide whether we go all-in PWA or try native iOS / Android, and then either drop this component or rework all the other places where we have <div>s and the alike. Currently im more in favor of PWA)

src/components/AsNavigationItem.tsx
```diff
@@ -6,6 +6,9 @@ type Props = {
     children: ReactNode | ReactNode[];
 };
 
+/**
+ * It allows to control the active navigation path by using a effect to set `rootPath`on initial rendering
+ */
```
Jonasdoubleyou, line 11: (hm, not sure if we really need a global state just to highlight the icon in the navigation bar. Maybe we can just move the navItems into the bar, and then just decide based on the current location which one to highlight)

Jonasdoubleyou (approved):
nice cleanup!

src/components/CSSWrapper.tsx
JeangelLF: Sounds good, I also think all-in PWA would be a nice idea.

Jonasdoubleyou (approved):

PR #498, opened 2024-03-15
feature: use new achievement text mappings

dhenkel92:
# Based on PR #496

### DIFF
--- a/src/components/achievements/modals/AchievementModal.tsx
+++ b/src/components/achievements/modals/AchievementModal.tsx
@@ -13,9 +13,11 @@ import { useLocation, useNavigate } from 'react-router-dom';
 import { Linking } from 'react-native';
 
 type AchievementModalProps = {
-    title?: string;
-    name: string;
+    tagline?: string;
+    title: string;
+    subtitle?: string;
     description: string;
+    footer?: string;
     achievementState: Achievement_State;
     achievementType: Achievement_Type_Enum;
     buttonText?: string;
@@ -24,8 +26,6 @@ type AchievementModalProps = {
     steps: Step[];
     maxSteps?: number;
     currentStep?: number;
-    progressDescription?: string;
-    achievedText?: string;
     image?: string;
     alternativeText?: string;
     onClose?: () => void;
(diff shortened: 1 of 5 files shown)


src/components/achievements/modals/AchievementModal.tsx
```diff
@@ -228,7 +226,7 @@ const AchievementModal: React.FC<AchievementModalProps> = ({
                                                         <IndicatorBar
                                                             maxSteps={maxSteps}
                                                             currentStep={currentStep}
-                                                            progressDescription={progressDescription}
+                                                            progressDescription={footer}
```
Jonasdoubleyou, line 229: rename also in this component?

src/config.ts
```diff
@@ -21,4 +21,4 @@ export const DD_ENV = window.liveConfig?.RUNTIME_DD_ENV ?? process.env.REACT_APP
 
 export const TALKJS_APP_ID = window.liveConfig?.RUNTIME_TALKJS_APP_ID ?? process.env.REACT_APP_TALKJS_APP_ID;
 
-export const GAMIFICATION_ACTIVE = (window.liveConfig?.RUNTIME_GAMIFICATION_ACTIVE ?? 'false') === 'true';
+export const GAMIFICATION_ACTIVE = (window.liveConfig?.RUNTIME_GAMIFICATION_ACTIVE ?? 'true') === 'true';
```
Jonasdoubleyou, line 24: Is "true" a good default? Maybe we should just activate in dev deployments / locally using the respective ENVs?

Jonasdoubleyou (approved):

src/config.ts
dhenkel92: 🙄 No it's not. Very good catch!

This was never supposed to be comitted.

src/components/achievements/modals/AchievementModal.tsx
dhenkel92: I would say that it's okay in this case. For an Indicator bar it's actually a progress description and we are just using the footer to save the text.

PR #499, opened 2024-03-16
Fix logic to add instructors to a course

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1163

## What was done?

- Adjusted the logic that adds instructors. Before, we were treating all the instructors we select on the "create mode" as if they were already stored in the database which didn't make sense and was causing some errors when trying to delete them from the course
- Removed unused state.

I tested it on both flows, create course and edit course and now we can remove instructors even if they're already stored in the database or if they're just part of our local state on the frontend.

https://github.com/corona-school/user-app/assets/161815068/e46c16c8-37e8-4d11-abed-b090760d1ea5

### DIFF
--- a/src/pages/CreateCourse.tsx
+++ b/src/pages/CreateCourse.tsx
@@ -841,8 +841,6 @@ const CreateCourse: React.FC = () => {
         setCurrentIndex(index);
     }, []);
 
-    console.log({ addedInstructors, newInstructors });
-
     return (
         <AsNavigationItem path="group">
             <WithNavigation
--- a/src/pages/CreateCourse.tsx
+++ b/src/pages/CreateCourse.tsx
@@ -116,7 +116,6 @@ const CreateCourse: React.FC = () => {
     const [image, setImage] = useState<string>('');
     const [courseAppointments, setCourseAppointments] = useState<Appointment[]>();
 
-    const [isPublished, setIsPublished] = useState(false);
     const [isLoading, setIsLoading] = useState<boolean>();
     const [showCourseError, setShowCourseError] = useState<boolean>();
 
@@ -336,7 +335,6 @@ const CreateCourse: React.FC = () => {
         setAllowParticipantContact(!!prefillCourse.allowChatContactParticipants);
         setAllowChatWriting(prefillCourse.groupChatType === ChatType.NORMAL ? true : false);
         setCourseClasses([prefillCourse.minGrade || 1, prefillCourse.maxGrade || 13]);
-        setIsPublished(prefillCourse.published ?? false);
         setCourseAppointments(prefillCourse.appointments ?? []);
         prefillCourse.course.image && setImage(prefillCourse.course.image);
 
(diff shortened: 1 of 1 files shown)


src/pages/CreateCourse.tsx
```diff
@@ -799,18 +797,14 @@ const CreateCourse: React.FC = () => {
...
-                }
+            const instructorExistsInArray = newInstructors.some((e) => e.id === instructor.id);
+            if (!instructorExistsInArray) {
+                // Instructors are only added "locally"
+                setNewInstructors((prev) => [...prev, instructor]);
             }
             hide();
         },
-        [addedInstructors, newInstructors, prefillCourseId, hide]
+        [newInstructors, hide]
```
Jonasdoubleyou, line 807: Can we remove addedInstructors completely then? Or is it used elsewhere?

Jonasdoubleyou: Also in `finishCreation` we read the `addedInstructors`, not the `newInstructors`. 

Maybe we should just maintain an array of "instructors", then when saving diff that with the course instructors and then add and delete (I don't see a reason why in the course edit we apply the change directly)

Jonasdoubleyou:
FYI: We somewhen decided to use https://www.conventionalcommits.org/en/v1.0.0/ (but we do not follow it consistently) ... So the proper title for the PR (which becomes the commit message when squashed and merged) would be "fix: ..."

src/pages/CreateCourse.tsx
JeangelLF: > Can we remove addedInstructors completely then? Or is it used elsewhere?

It's being used in a few other places unfortunately and I wanted to avoid a lot of refactor on this task.

> Maybe we should just maintain an array of "instructors", then when saving diff that with the course instructors and then add and delete (I don't see a reason why in the course edit we apply the change directly)

Yeah, something like that should be the way. Recently I assigned to myself [this task you created](https://github.com/corona-school/project-user/issues/879) to refactor this whole Create/Edit course stuff. It's quite a mess so it's going to be fun.

JeangelLF:
> FYI: We somewhen decided to use https://www.conventionalcommits.org/en/v1.0.0/ (but we do not follow it consistently) ... So the proper title for the PR (which becomes the commit message when squashed and merged) would be "fix: ..."
Thanks, I adjusted it. Do we also have a convention for branch naming?

Jonasdoubleyou:
> Do we also have a convention for branch naming?

No, they are deleted anyways upon merge, the PR name will stay forever in the git history

src/pages/CreateCourse.tsx
Jonasdoubleyou: Okay, we can move that refactoring to somewhen, but my second comment is still relevant: If you create a new course, will it properly pick up the added instructors (`finishCreation` using `addedInstructors`)?

JeangelLF: Sure, it's still relevant, I'll take a look. 
After adjusting that, we'll be using `addedInstructors` only to display them [here](https://github.com/corona-school/user-app/blob/2512d51fc820403154790749c4c193cafdbe64ba/src/pages/course-creation/FurtherInstructors.tsx#L65) and to remove them with a call to the API [here](https://github.com/corona-school/user-app/blob/e8801dc30a0f60a1a87936d6fd30ab6e1bcf1f4d/src/pages/CreateCourse.tsx#L817).

It's referenced in another component but is unused.

JeangelLF: Ok, I've adjusted it to read `newInstructors` which fix the issue you mentioned. 
I tested it and it takes the instructors correctly while creating/editing. It also allows removal of instructors during both flows.

Jonasdoubleyou (approved):

PR #501, opened 2024-03-17
feature: hide progres bar for future/completed achievements

dhenkel92:
# Based on PR #500

This will remove the progress bar from future achievements like this

image

### DIFF
--- a/src/components/achievements/achievementCard/AchievementCard.tsx
+++ b/src/components/achievements/achievementCard/AchievementCard.tsx
@@ -20,6 +20,7 @@ type AchievementCardProps = {
     maxSteps?: number;
     currentStep?: number;
     progressDescription?: string;
+    showProgressBar: boolean;
     onClick?: () => void;
 };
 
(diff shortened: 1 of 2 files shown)


src/components/achievements/achievementCard/AchievementCard.tsx
```diff
@@ -140,7 +142,7 @@ const AchievementCard: React.FC<AchievementCardProps> = ({
                                 {title}
                             </Text>
                         </Stack>
-                        {achievementState !== Achievement_State.Completed && (
+                        {achievementState !== Achievement_State.Completed && showProgressBar && (
```
Jonasdoubleyou, line 145: !== completed is now redundant?

Jonasdoubleyou (approved):

src/components/achievements/achievementCard/AchievementCard.tsx
dhenkel92: Ah, right. I'll remove it with the next PR

PR #508, opened 2024-03-26
Feat: Add settings on the screening blocker

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1159

## What was done?

- Added a new `RequireScreeningSettingsDropwon` that will appear on the screening blocker. This component allows the user to deactivate their account, or change their notifications preferences _(via a new `NotificationsPreferencesModal` which just re-uses the existing component but in a modal)_.
- Added links to Datenschutz and Impressum, similar to how we do it in the login page
- Slightly updated how a "full" modal look like, so it at least have some space on the sides which looks a bit better. I checked and this modal size was only being used on the screeners section, and it's still looking good with the new changes.

Here is how it looks like:

https://github.com/corona-school/user-app/assets/161815068/22bd1564-7727-433b-810d-24ba4939fd05

### DIFF
--- a/src/components/Tabs.tsx
+++ b/src/components/Tabs.tsx
@@ -14,9 +14,10 @@ type Props = {
     removeSpace?: boolean;
     onPressTab?: (tab: Tab, index: number) => any;
     tabInset?: number | string;
+    shouldRedirect?: false;
 };
 
-const Tabs: React.FC<Props> = ({ tabs, removeSpace = false, onPressTab, tabInset }) => {
+const Tabs: React.FC<Props> = ({ tabs, removeSpace = false, onPressTab, tabInset, shouldRedirect = true }) => {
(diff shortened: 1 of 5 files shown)


src/components/Tabs.tsx
```diff
@@ -32,7 +33,9 @@ const Tabs: React.FC<Props> = ({ tabs, removeSpace = false, onPressTab, tabInset
     const Tab = ({ tab, index, active }: { tab: Tab; index: number; active: boolean }) => (
         <Pressable
             onPress={() => {
-                navigate('', { replace: true, state: { tabID: index } });
+                if (shouldRedirect) {
```
JeangelLF, line 36: Kind of a temporary hack to be able to use the Tabs inside of a modal. In a separate task, I'll refactor this component a bit to split it into two so it's easier to use.
https://github.com/corona-school/project-user/issues/1169

Jonasdoubleyou (approved):
nice!

PR #509, opened 2024-03-27
Refactor: Split Tabs component into two

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1169

## What was done?

- Extracted Navigation Logic out of the Tabs component. Now, we have a new `NavigationTabs` which contains all that logic and `Tabs` is quite basic now.
- Updated imports that were using `Tabs` to use `NavigationTabs`. Both components have the same contract, so no usage changes
- Removed the `shouldRedirect` hack I introduced last time

### DIFF
--- a/src/components/NavigationTabs.tsx
+++ b/src/components/NavigationTabs.tsx
@@ -1,0 +1,30 @@
+import { useEffect } from 'react';
+import { useLocation, useNavigate } from 'react-router-dom';
+import { useRestoredNumberState } from '../hooks/useScrollRestoration';
+import Tabs, { type Tab as BaseTab, type TabsProps } from './Tabs';
+
+export type Tab = BaseTab;
+
+type NaivgationTabsProps = TabsProps & {};
+
+const NavigationTabs: React.FC<NaivgationTabsProps> = ({ tabs, removeSpace = false, onPressTab, tabInset }) => {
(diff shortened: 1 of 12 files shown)


src/components/NavigationTabs.tsx
```diff
@@ -0,0 +1,30 @@
+import { useEffect } from 'react';
+import { useLocation, useNavigate } from 'react-router-dom';
+import { useRestoredNumberState } from '../hooks/useScrollRestoration';
+import Tabs, { type Tab as BaseTab, type TabsProps } from './Tabs';
+
+export type Tab = BaseTab;
+
+type NaivgationTabsProps = TabsProps & {};
```
realmayus, line 8: What does `& {}` do?

JeangelLF: That's for [extending types](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces). I was using it to extend the types of `TabsProps` as I had some extra props on this component.
```ts
type NaivgationTabsProps = TabsProps & {
  somePropThatIsNotInTabsProps: string
};
```
I forgot to remove it after realizing I wasn't going to need more props.
Ps: I just noticed the typo on `NaivgationTabsProps`  🙈

realmayus (approved):

PR #510, opened 2024-03-28
feat: Add a 'new' badge to newly published courses

josefacastillo:
info: I still need to get publichedAt into SubcourseData, didnt have more time

Update: done :P


Jonasdoubleyou (approved):
nice! :)

src/widgets/AppointmentCard.tsx
```diff
@@ -50,6 +51,7 @@ type Props = {
     description: string;
     maxParticipants?: number;
     participantsCount?: number;
+    publishedAt?: Date;
```
Jonasdoubleyou, line 54: As it is an optional property, do we fetch it in all relevant GraphQL queries?

josefacastillo: Good question. I wasnt sure about that actually, how can I find out if is always there?

Jonasdoubleyou: You could make it mandatory and see if it still compiles :)

Jonasdoubleyou: Ah, nevermind - it's also nullable in the model. Not sure but : Date | null might work (excluding undefined)

Jonasdoubleyou: Hm, for nullable properties that is tricky as we mix null and undefined ... I think it is fine if you check (Dashboard + Group Page) x (Pupil + Student) :)

Jonasdoubleyou (approved):

PR #511, opened 2024-03-28
feat: Mandatory screening for students

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1154

## What was done

- Updated existing mandatory screening modal for pupils to also work for students
- Included some very minor refactor about the creation of screening links
- Add a default selected tab value in the `Tabs` component

Here is how it looks like:
Bildschirmfoto 2024-03-28 um 11 41 28
Bildschirmfoto 2024-03-28 um 11 52 40

## Note

During my tests I only noticed a weird behavior and is that when approving a student for tutoring, their sessions isn't being updated, so they cannot continue to the app until they re-login.
I'll follow up with a backend PR to update [here](https://github.com/corona-school/backend/blob/1c73699ee9ad757187e7f90b29ffe4e579a71747/common/student/screening.ts#L64) the user session when they get approved for tutoring. Similar to how we are doing it [here](https://github.com/corona-school/backend/blob/1c73699ee9ad757187e7f90b29ffe4e579a71747/common/student/screening.ts#L40) for instructors

### DIFF
--- a/src/modals/RequireScreeningModal.tsx
+++ b/src/modals/RequireScreeningModal.tsx
@@ -6,9 +6,11 @@ import LokiIcon from '../assets/icons/lernfair/avatar_pupil_120.svg';
 import { useTranslation } from 'react-i18next';
 import { gql } from '../gql';
 import { useQuery } from '@apollo/client';
-import useApollo from '../hooks/useApollo';
+import useApollo, { useUserType } from '../hooks/useApollo';
 import CenterLoadingSpinner from '../components/CenterLoadingSpinner';
 import RequireScreeningSettingsDropdown from '../widgets/RequireScreeningSettingsDropdown';
+import { asTranslationKey } from '../helper/string-helper';
+import { createPupilScreeningLink, createStudentScreeningLink } from '../helper/screening-helper';
 
 const EXISTING_SCREENINGS_QUERY = gql(`  
     query ExistingScreenings {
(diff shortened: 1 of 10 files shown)


src/modals/RequireScreeningModal.tsx
```diff
@@ -53,56 +76,47 @@ export function RequireScreeningModal() {
...
             </HStack>
             {!data && <CenterLoadingSpinner />}
-            {data && !wasScreened && !wasRejected && (
+            {data && needScreening() && (
                 <VStack maxW={sizes['smallWidth']} space={space['1']} flex="1" alignItems="center">
                     <Spacer />
                     <EventIcon />
                     <Heading size="md" textAlign="center" color="lightText">
-                        {t('requireScreening.noScreening.title', { firstname: user?.firstname })}
+                        {t(asTranslationKey(`requireScreening.${userType}.noScreening.title`), { firstname: user?.firstname })}
```
Jonasdoubleyou, line 84: I think alternatively you could fix the return type of useUserType to return "pupil" | "student" | "screener" instead of string, then it should work correctly here...

Jonasdoubleyou (approved):

realmayus (approved):

src/modals/RequireScreeningModal.tsx
JeangelLF: The return type of useUserType is already inferred to `"pupil" | "student" | "screener"`
The problem here is probably that it includes the "screener" part which doesn't have a corresponding translation key, and that makes sense in this case.

PR #533, opened 2024-05-06
feat: Cache GraphQL queries

Jonasdoubleyou:
Basic idea: Keep a cache for each query's result, if a query is fetched again, directly return the value from cache but also refetch the data in the background, then update the cache and the page with that. This way, the app more or less stays consistent and stays "as is" on devices with low-latency. On a bad connection (or even offline) one can still navigate around the app and the old data is displayed. 

Screenshot 2024-05-09 at 00 27 06


Jonasdoubleyou:
Do we want to go down this route? We could either (a) use Apollo's InMemoryCache, but then we need to adapt all queries to use `returnPartialData: true` (and some other config I do not find that intuitive), or (b) we implement our own "result cache" (outlined something in the current version)

JeangelLF:
In what scenarios isn't the cache working right now? I haven't tested it, just curious to know if it isn't working at all or just partially.

> We could either (a) use Apollo's InMemoryCache, but then we need to adapt all queries to use returnPartialData: true

Hmm, I don't like this and I don't understand why it works with that either. 

I'd say Let's go with b. Should we [persist the cache](https://www.apollographql.com/docs/react/caching/advanced-topics/#persisting-the-cache) in some sort of storage? 🤔  Not sure if we needed it since our navigation is client side.

Jonasdoubleyou:
Regarding persistence: I guess for now that creates more potential problems (privacy) than it solves things - as the App really won't work offline, so at some point we anyways need to fetch from the server

JeangelLF (approved):
🚀

PR #539, opened 2024-05-16
refactor if- to switch-statement in courseState info text

Fabian5150:
Refactor if- to switch-statement in courseState info text


Fabian5150:
Had this laying around in another branch that I ended up deleting, but wanted to push this little part of it.

realmayus (approved):

PR #540, opened 2024-05-16
Allow editing/deleting of unpublished course's appointments

Fabian5150:
Allowing students to edit & delete appointments of their unpublished courses


Fabian5150:
Closes https://github.com/corona-school/project-user/issues/1198 (for now, the mentioned UX changes will be part of a bigger change in the future)

Fabian5150:
Does not include the deleting of past lectures. 
Originally I wanted to do both things at once, but the deleting of past lectures turns out to be a bit more than anticipated, so I didn't want this feature to have to wait for that.

Jonasdoubleyou (approved):

PR #542, opened 2024-05-21
feat: Service Worker + Web Push

Jonasdoubleyou:
If the runtime config `RUNTIME_SERVICE_WORKER_ACTIVE` is turned on, registers a service worker that
- Caches static resources of the App (with a bit of config the app would start offline)
- Receives Push Notifications, and shows a generic notification
- Handles a click on the notification and opens the User App

Also adds a hidden page `/push` where one can turn on push notifications - might come in handy during a "Beta" phase were we ask users to test it (very rudimentary UI). 

https://github.com/corona-school/project-user/issues/606

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -14,6 +14,7 @@
       "react-hooks/exhaustive-deps": "warn",
       "no-unused-expressions": "off",
       "lernfair-app-linter/typed-gql": "warn",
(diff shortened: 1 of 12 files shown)


JeangelLF (approved):
Cool! 🚀

PR #555, opened 2024-06-05
feat: Add safe area on BottomNavigationBar

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1226

## What was done?

- Added safe area padding on the `bottomNavigationBar`. It should only affect devices with a swipe bar

Bildschirmfoto 2024-06-05 um 09 28 16
Bildschirmfoto 2024-06-05 um 09 18 34

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -3,7 +3,7 @@
     <head>
         <meta charset="utf-8" />
         <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
-        <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1" />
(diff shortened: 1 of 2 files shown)


Jonasdoubleyou (approved):

Jonasdoubleyou:
What about the "dynamic island" thingy? I remember spending hours with bugs on this ...

JeangelLF:
> What about the "dynamic island" thingy? I remember spending hours with bugs on this ...

Bildschirmfoto 2024-06-05 um 09 54 15

Oh, interesting. I didn't know about that. I just tried the app with a device that has that and it seems to be ok _(I think)_. It's also worth noting that I don't know how to use that "feature".

PR #560, opened 2024-06-06
Feat/Fix: Add Prospects to Subcourse

realmayus:
- Add prospect list in course overview to allow instructors to add prospects (frontend part of https://github.com/corona-school/backend/pull/1093)
- This reuses the waiting list component
- Update browserslist


src/pages/single-course/ProspectList.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+    const handleOpenModal = (prospect: SparseParticipant) => {
+        setIsJoinPupilModalOpen(true);
+        setPupilToAdd(prospect);
+    };
+
+    const handleAddPupil = useCallback(
+        async (pupilId: number) => {
+            try {
+                await addProspect({ variables: { subcourseId: subcourseId, pupilId: pupilId } });
```
Jonasdoubleyou, line 41: while this is loading I could click the button again

src/pages/ChangeEmail.tsx
```diff
@@ -84,7 +84,11 @@ const ChangeEmail = () => {
                         <Row justifyContent="center">
                             <DisableableButton
                                 isDisabled={!canChange}
-                                reasonDisabled={t('reasonsDisabled.invalidEMail')}
+                                reasonDisabled={
+                                    data?.me?.email.toLowerCase().trim() === newEmail?.toLowerCase().trim()
+                                        ? t('reasonsDisabled.didNotChangeEmail')
+                                        : t('reasonsDisabled.invalidEMail')
+                                }
                                 width={buttonWidth}
```
Jonasdoubleyou, line 92: How does this "Prevent invalid email changes" - It just swaps the disabled message in that case?

Jonasdoubleyou: And if it hits the uniqueness constraint we probably also miss validation on backend side.

realmayus: will move this to https://github.com/corona-school/project-user/issues/1349

JeangelLF (dismissed):
(_This can be tackled in a separate task_) I noticed a small, unexpected behavior:

When a pupil clicks on `Kursleiter:in kontaktieren`, they are immediately marked as a prospect on the CoursePage, even if they haven't sent a message yet. This feels a bit unexpected (At least for me 😅)

Could we:

- Filter out empty chat prospects: Only show users as prospects once they've sent a message.
- Add a confirmation modal: Inform the user that they’ll be added as a prospect after confirming.
----------

Other than that, it looks good, thank you! 💯

src/pages/student/SingleCourseStudent.tsx
```diff
@@ -407,16 +420,31 @@ const SingleCourseStudent = () => {
...
                                         refetchInstructorData();
                                         return refetchBasics();
                                     }}
+                                    type={'waitinglist'}
+                                />
+                            </TabsContent>
+                        )}
+                        {showWaitingListProspectListTab && (
+                            <TabsContent value="prospect-list">
+                                <WaitingListProspectList
```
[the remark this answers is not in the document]
realmayus: Nice idea! I'll implement it here.

realmayus:
How could we implement filtering out the empty chats?
Either, we could 
- fetch the chat data from TalkJS on every call of the resolver, or
- the backend could set a flag in the prospect chat DB entity
  - for that, the backend would have to be notified when the first message was sent in a chat (possibly via a TalkJS webhook?). 

I think the latter should be our preferred approach for performance reasons.
What do you think? Did I miss any more obvious solutions?

JeangelLF:
> he backend could set a flag in the prospect chat DB entity
for that, the backend would have to be notified when the first message was sent in a chat (possibly via a TalkJS webhook?).

That one makes sense. But now that I think about it twice, even if we filter out empty chats, the flow still would be a bit weird. It would be something like:
```
1. SuS clicks on Kursleiter kontaktieren
2. SuS sends whatever message on the chat (under the hood, this already marks the user as a prospect)
3. HuH adds the user to the course
4. SuS gets a notification that they joined the course (???)
```

This feels a bit off. I think we should just add a modal informing the user about what happens if they proceed to the chat.
What do you think? 👀

realmayus:
I think this does make sense, as HuH know that prospects are interested users that contacted them via chat. We should change the notfication action though, good idea. Then we can have something like "Max has added you to course X"

JeangelLF (approved):
Thank you!

PR #562, opened 2024-06-09
feat: Gradual rollout of new features

Jonasdoubleyou:
Allow features to be gradually enabled for a random subset of users, which can gradually be increased.

Participants are tracked via Datadog RUM, to be able to identify issues with new features.

resolves https://github.com/corona-school/project-user/issues/1238

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,3 +1,30 @@
+import { datadogRum } from '@datadog/browser-rum';
+
+// For each device that opens the App, pick a random number [0, 100) and store it in the local storage
+// If that number is smaller than the configured 'rollout ratio' the device is chosen to participate in the beta
+//
+// Thus:
+// - If a device was once picked to be in the beta, it stays in the beta (unless the rollout ratio is decreased)
+// - The rollout ratio can be gradually increased, i.e. from 10% to 20%
+// - A random set of users is chosen per feature toggle
+// - Feature toggles are per device, not per user (as doing this per user would be much more complicated)
+function betaRollout(name: string, rolloutRatioString: string | undefined) {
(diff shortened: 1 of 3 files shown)


JeangelLF (approved):
Very interesting, thank you!

src/config.ts
```diff
@@ -1,3 +1,30 @@
...
+    let randomString = localStorage.getItem(key);
+    if (!randomString) {
+        randomString = Math.floor(Math.random() * 100).toString();
+        localStorage.setItem(key, randomString);
+    }
+
+    let random = parseInt(randomString, 10);
+    let rolloutRatio = parseInt(rolloutRatioString ?? '0', 10);
+
+    const chosen = random < rolloutRatio;
```
dhenkel92, line 22: This will be false if random is 100, should be `random <= rolloutRatio`

image

dhenkel92 (commented):
Pretty cool 🙂 👍

src/config.ts
Jonasdoubleyou: Argh, floating point math. Math.random() is [0, 1), so *100 should be [0, 100) but I guess it is actually [0, 100 + Epsilon) and 100 + Epsilon might not be < 100 ...

Jonasdoubleyou: On the contrary, if the rollout ratio is 0, <= would also be wrong ... I guess we should just turn the rollout ratio to 101, or just remove the toggle if we turn it on for all users anyways

dhenkel92: That's true. What about being more explicit in code? We could set chose to true if rolloutRation is 100% and to false if it's zero. It's a couple more lines of code, but ensures that these boundaries are always given, no matter what `random` is.

PR #570, opened 2024-06-18
feat: show that in summer vacation no Hausaufgabenhilfe is available

MartinEichinger:
Dashboard.tsx: Introduction of a conditional selection of the existing CTA or a text block with the holiday notice. Introduction of a boolean variable to query whether the current date is within the vacation period.

barrier-block_green.svg: new image used for the vacation notice

de.json: introduced new matching.homeworkhelp.buttonSummerVacation to define text for vacation notice

Resolves corona-school/project-user#1240


JeangelLF (commented):
Thanks a lot for the PR! 🚀

Just a few minor suggestions:
1. On desktop, can we make the banner appear below the text? `Jeden tag von Montag bis...`
Something like:
Bildschirmfoto 2024-06-19 um 15 09 25

2. Let's remove the `...` from the text. It was like that in the mockup but we just realized it gives the feeling that there is more text but it's just hidden behind an ellipsis.

3. The banner color is not correct but this is because our Theme is not configured correctly :(
Could you change on the `Theme.ts` file the value of `secondary.100` to be something like: `#FBEFC6`

MartinEichinger:
Thanks for the feedback. I just updated according to your review feedback:
1) Changed the ```min-width``` of the banner to ```100%```
2) Removed all ```...``` from the translation files
3) Introduced ```VAR_COLOR_YELLOW_LIGHT = '#FBEFC6'``` and assigned it to ```secondary.100```

JeangelLF (approved):
Cool, thank you! :shipit:

JeangelLF (approved):

PR #571, opened 2024-06-18
Feat: Promote App Installation

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1222

## What was done?

- Added banners to promote the app installation
- Added a context to handle the app installation _(if available)_
- Added a button on the settings page to install the app (if possible / needed)
Bildschirmfoto 2024-06-18 um 16 54 39
Bildschirmfoto 2024-06-18 um 16 53 59


src/widgets/InstallAppBanner.tsx
```diff
@@ -0,0 +1,77 @@
...
+import IconClose from '../assets/icons/ic_close.svg';
+import IconShare from '../assets/icons/icon_share.svg';
+import IconAdd from '../assets/icons/icon_add_square.svg';
+import { useLocalStorage } from '../hooks/useLocalStorage';
+import { Trans, useTranslation } from 'react-i18next';
+
+export const NativeInstallAppBanner = () => {
+    const { shouldPromote, promote, promotionType } = useInstallation();
+    const { t } = useTranslation();
+    const [show, setShow] = useLocalStorage({ key: 'recommend-lern-fair-installation', initialValue: true });
```
Jonasdoubleyou, line 12: (maybe we could just determine the value of show in this component, and move all the other hooks into a child component. Then we would short circuit here for most users - after all this is added to every page through the nav bar - not sure if the impact is measurable but I would try to keep the main renders as small as possible)

Jonasdoubleyou (approved):
nice!

src/widgets/InstallAppBanner.tsx
JeangelLF: I actually moved the `useInstallation` logic to a context since I needed to capture and keep the `beforeinstallprompt` event to be able to install the app via the banner / settings page.

The "good" thing with this in terms of performance is that users will see the banner probably once and that would be it.

src/pages/Settings.tsx
```diff
@@ -126,6 +158,9 @@ const Settings: React.FC = () => {
                         </Column>
                     </ProfileSettingRow>
                 </VStack>
+                {showInstallInstructions && promotionType === PromotionType.iPhone && (
+                    <IOSInstallAppInstructions onClose={handleOnCloseInstallInstructions} variant={'iPhone'} />
+                )}
```
Jonasdoubleyou, line 163: Can't the context show these as a fixed overlay?

Jonasdoubleyou (dismissed):

src/pages/Settings.tsx
JeangelLF: Thanks for the suggestions! You're right. I moved those components to the context

Jonasdoubleyou (approved):

PR #572, opened 2024-06-18
fix: Remove query from navigation bars

Jonasdoubleyou:
No need to query the user roles manually if they are available via context ...

This significantly removes flicker when navigating across the App (as the navigation bar does not need to rerender, thus causing a layout shift)


JeangelLF (approved):
Thanks for the fix! It looks great!!! 🤩

PR #573, opened 2024-06-20
fix: delete count in AppointmentForm

MartinEichinger:
InputSuffix.tsx: set appointmentsCount as optional input parameter. Name of input field is named "Lektion" if no appointmentsCount is provided and "Lektion #{appoinmentsCount}" otherwise.

AppoitmentForm.tsx: delete input prop appointmentsCount from <Input Suffix />. AppointmentForm.tsx is only used once and appointmentsCount is also no longer used in AppointmentsForm, so I deleted the input prop appointmentsCount.

AppointmentCreation.tsx: deleted input prop appointmentsCount from <AppointmentForm />.

Resolves corona-school/project-user#1252

### DIFF
--- a/src/pages/create-appointment/AppointmentForm.tsx
+++ b/src/pages/create-appointment/AppointmentForm.tsx
@@ -21,16 +21,7 @@ type FormProps = {
     videoChatType: string;
     isCourse: boolean;
 };
-const AppointmentForm: React.FC<FormProps> = ({
-    errors,
-    // appointmentsCount,
-    onSetDate,
-    overrideMeetingLink,
-    onSetTime,
-    isCourse,
-    setVideoChatType,
-    videoChatType,
-}) => {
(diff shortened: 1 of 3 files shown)


src/pages/create-appointment/AppointmentForm.tsx
```diff
@@ -24,7 +23,7 @@ type FormProps = {
 };
 const AppointmentForm: React.FC<FormProps> = ({
     errors,
-    appointmentsCount,
+    // appointmentsCount,
```
Jonasdoubleyou, line 26: remove

Jonasdoubleyou (dismissed):

src/pages/create-appointment/AppointmentForm.tsx
MartinEichinger: Many thanks for the hint. I removed it.

Jonasdoubleyou (approved):

PR #574, opened 2024-06-20
Feat: Add /logout route, redirect there from logout button

realmayus:
Add /logout route, redirect there from logout button

Closes https://github.com/corona-school/project-user/issues/1234


src/components/Logout.tsx
```diff
@@ -0,0 +1,11 @@
+import useApollo from '../hooks/useApollo';
+import { useNavigate } from 'react-router-dom';
+
+export default function Logout() {
+    const navigate = useNavigate();
+    const { logout } = useApollo();
+    logout();
```
Jonasdoubleyou, line 7: It's a Promise, better wait for it :)

src/components/Logout.tsx
```diff
@@ -0,0 +1,36 @@
...
+            } catch (error) {
+                logError('Authentication', 'Failed to logout', error);
+            }
+        }, []);
+    };
+
+    const logout = useLogout();
+    logout().then(() => navigate('/welcome', { state: locState }));
+
+    return null;
```
Jonasdoubleyou, line 35: Maybe a loading spinner instead of an empty page?

src/components/Logout.tsx
```diff
@@ -0,0 +1,36 @@
...
+            try {
+                await logout();
+            } catch (error) {
+                logError('Authentication', 'Failed to logout', error);
+            }
+        }, []);
+    };
+
+    const logout = useLogout();
+    logout().then(() => navigate('/welcome', { state: locState }));
```
Jonasdoubleyou, line 33: If we login again, on which page will we land?

realmayus: We land on /start

Jonasdoubleyou (approved):

PR #581, opened 2024-06-24
feat: Add language to user chat options

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1231

## What was done?

 Added language to the user options that we pass to talkJS

### DIFF
--- a/src/context/ChatContext.tsx
+++ b/src/context/ChatContext.tsx
@@ -6,6 +6,7 @@ import { useQuery } from '@apollo/client';
 import { UnreadConversation } from 'talkjs/all';
 import { userIdToTalkJsId } from '../helper/chat-helper';
 import { TALKJS_APP_ID } from '../config';
+import { useTranslation } from 'react-i18next';
 
 interface UserType {
     userID: string;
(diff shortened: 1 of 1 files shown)


src/context/ChatContext.tsx
```diff
@@ -95,12 +97,12 @@ export const LFChatProvider: React.FC<{ children: ReactNode }> = ({ children })
...
         // Wait for the User's Chat Signature to be fetched from the Backend:
         if (loading) return;
-
-        const me = {
+        const me: Talk.UserOptions = {
             id: userIdToTalkJsId(userId),
             name: getChatName(user),
             role: user?.pupil ? 'pupil' : 'student',
             email: user?.email,
+            locale: i18n.language,
```
Jonasdoubleyou, line 105: Are we sure that locale = language in all cases?

Jonasdoubleyou (approved):

src/context/ChatContext.tsx
JeangelLF: Yeah, I tried it on my local with the languages we offer and it looks good. I'd like to try it on STG and see how it goes because the chat on my local is different. I'll adjust it if it's not correct for a language.

PR #591, opened 2024-06-28
feat: group subjects in subject selector

MartinEichinger:
The subject selector was divided into 3 groups - main subjects, minor subjects and rare subjects.
Subjects in low demand have been removed from the display and 4 new subjects have been added instead.
New required icons have been created.
The height of the subject cards within a row has been adjusted.

Resolves corona-School/project-user#1043


src/lang/ar.json
```diff
@@ -1,2329 +1,2385 @@
 {
-    "appointment": {
```
Jonasdoubleyou, line 2: Did you reformat the language files manually? We should avoid large diffs in these files.

MartinEichinger: Maybe I generate the files again with the newest from main.

MartinEichinger: I'm not sure, whether this is relevant, but I do not understand following: 
- the language files from main have 4 whitespaces as identation. 
- when I generate the files with the translate script, they only have 2 whitespaces as identation.

MartinEichinger: @Jonasdoubleyou I found out, that the problem was the different identation (4 vs 2). I corrected the files and committed them again. Doing the review on all commits should show now a clear diff file.

src/widgets/SubjectSelectorGrouped.tsx
```diff
@@ -0,0 +1,140 @@
...
+    groupWidth,
+    justifyContent,
+    limit,
+    subjects,
+    addSubject,
+    removeSubject,
+    variant,
+    importance = 0,
+}: {
+    subSubjects: Subject['name'][];
```
Jonasdoubleyou, line 20: (Maybe rather selectableSubjects? And instead of "subjects" chosenSubjects?)

src/widgets/SubjectSelectorGrouped.tsx
```diff
@@ -0,0 +1,140 @@
...
+    addSubject,
+    removeSubject,
+    limit,
+    selectable,
+    variant = 'selection',
+    includeDaz = false,
+    justifyContent,
+}: {
+    subjects: Subject['name'][];
+    selectable?: Subject['name'][];
```
Jonasdoubleyou, line 85: unused?

src/types/subject.ts
```diff
@@ -3,15 +3,18 @@ import { Course_Subject_Enum, Subject } from '../gql/graphql';
 // Keep in sync with Backend!
 export const SUBJECTS = [
```
Jonasdoubleyou, line 4: (We could maybe introduce `type Subject = (typeof SUBJECTS)[number]` and then use that instead of "string")

Jonasdoubleyou (dismissed):
nice, some minor improvements that could be made. 

Does the new layout also work well on different device sizes?

src/lang/ar.json
Jonasdoubleyou: hm, interesting ... not sure where that difference in formatting came from

src/widgets/SubjectSelectorGrouped.tsx
MartinEichinger: Right. I delete it.

src/widgets/SubjectSelectorGrouped.tsx
MartinEichinger: Sounds good. I will change it.

MartinEichinger:
> nice, some minor improvements that could be made.
> 
> Does the new layout also work well on different device sizes?

@Jonasdoubleyou I would say so. I checked with the Chrome Screensizes "iPhone SE", "iPhone 12 Pro", "iPad Air" and "iPad Pro".

src/types/subject.ts
MartinEichinger: Good idea. Only the type Subject already exists in gql/graphql. I would take type SingleSubject = typeof SUBJECTS[number]

src/widgets/SubjectSelectorGrouped.tsx
```diff
@@ -0,0 +1,136 @@
...
+    variant,
+    importance = 0,
+}: {
+    selectableSubjects: Subject['name'][];
+    group: string;
+    groupWidth: string;
+    justifyContent?: string;
+    limit?: number;
+    chosenSubjects: Subject['name'][];
+    addSubject: (name: string) => void;
```
Jonasdoubleyou, line 26: (could then also use SingleSubject in this component)

Jonasdoubleyou (dismissed):

src/widgets/SubjectSelectorGrouped.tsx
MartinEichinger: ok. I use it for selectableSubjects and chosenSubjects and also for subjects in Subjects.tsx.

Jonasdoubleyou (approved):

PR #593, opened 2024-07-02
Fix: Push subscribe and unsubscribe behavior when logout/login

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1264

## What was done?

- Split the unsubscribe process into two, giving "priority" to the local unsubscription
- Store the user decision of the Switch (on/off) on local storage _(Not sure why I didn't do this in the first place 🤦 )_
- Unsubscribe when a user logs out
- Subscribe if user preferences require us to do so and if webPush state is `not-subscribed`

### DIFF
--- a/src/context/WebPushProvider.tsx
+++ b/src/context/WebPushProvider.tsx
@@ -21,12 +21,14 @@ const WebPushProvider = ({ children }: { children: React.ReactNode }) => {
     const { sessionState } = useUserAuth();
     const { status, subscribe, unsubscribe } = useWebPush();
     useEffect(() => {
-        if (!WEBPUSH_ACTIVE) return;
-        if (sessionState === 'logged-out') {
-            unsubscribe();
-        } else if (sessionState === 'logged-in' && status === 'not-subscribed' && pushEnabled) {
-            subscribe();
-        }
+        (async () => {
+            if (!WEBPUSH_ACTIVE) return;
+            if (sessionState === 'logged-out') {
+                await unsubscribe();
+            } else if (sessionState === 'logged-in' && status === 'not-subscribed' && pushEnabled) {
+                await subscribe();
+            }
+        })();
     }, [sessionState]);
 
     return <WebPushContext.Provider value={{ status, subscribe, unsubscribe }}>{children}</WebPushContext.Provider>;
--- a/src/context/WebPushProvider.tsx
+++ b/src/context/WebPushProvider.tsx
@@ -23,9 +23,7 @@ const WebPushProvider = ({ children }: { children: React.ReactNode }) => {
     useEffect(() => {
         (async () => {
             if (!WEBPUSH_ACTIVE) return;
-            if (sessionState === 'logged-out') {
-                await unsubscribe();
-            } else if (sessionState === 'logged-in' && status === 'not-subscribed' && pushEnabled) {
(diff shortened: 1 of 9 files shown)


src/lib/WebPush.ts
```diff
@@ -259,6 +256,18 @@ export function useWebPush() {
...
+        if (subId === null) return;
+
+        try {
+            await unsubscribeOnServer(client, subId);
+            setSubId(null);
+
+            setStatus('not-subscribed');
+            log('WebPush', 'WebPush subscription removed');
+        } catch (error) {
+            log('WebPush', 'Failed to remove subscription', error);
```
Jonasdoubleyou, line 269: Nobody will read this log, maybe somehow trace as error? (Not sure how to do this with Datadog)

src/context/WebPushProvider.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+const WebPushProvider = ({ children }: { children: React.ReactNode }) => {
+    const [pushEnabled] = useLocalStorage({ key: 'lern-fair-web-push-enabled', initialValue: false });
+    const { sessionState } = useUserAuth();
+    const { status, subscribe, unsubscribe } = useWebPush();
+    useEffect(() => {
+        (async () => {
+            if (!WEBPUSH_ACTIVE) return;
+            if (sessionState === 'logged-out') {
+                await unsubscribe();
```
Jonasdoubleyou, line 27: Hmm, not sure if doing this "after the logout" is robust enough ... 

Long term maybe we could introduce a DeviceID, and write that into secret + push subscription. Then we could directly detect logout / unused devices in the backend and remove push subscriptions.

Jonasdoubleyou (approved):

src/context/WebPushProvider.tsx
JeangelLF: You're right, it was just "working" because we invalidated the subscription, but the backend part always failed. The backend subscription was still deleted when we tried to use it... But in general, that wasn't right.

I tried to change it a bit to successfully remove the subscription from the backend as well just before the logout... It required a bit of "refactoring" to accomplish it but now it's working correctly.

> Long term maybe we could introduce a DeviceID, and write that into secret + push subscription. Then we could directly detect logout / unused devices in the backend and remove push subscriptions.

I agree that this can be improved with that.
I still feel there may be cases that I'm not covering with my current approach

src/lib/WebPush.ts
JeangelLF: I think it was being sent to Datadog with `datadogRum.addAction(line);` but not as an error, I added a separate function to do it with `datadogRum.addError(line, context);`

src/lib/WebPush.ts
```diff
@@ -243,23 +250,30 @@ export function useWebPush() {
...
+            await unsubscribeOnServer(client, subId);
+            setSubId(null);
+
+            setStatus('not-subscribed');
+            log('WebPush', 'WebPush subscription removed');
+        } catch (error) {
+            logError('WebPush', 'Failed to remove subscription', error);
+        }
+    }
+    console.log({ subId });
```
Jonasdoubleyou, line 277: ...

Jonasdoubleyou (approved):
nice!

PR #594, opened 2024-07-03
feat: Hide disabled inApp notifications

JeangelLF:
## What was done?

Hide disabled inApp notifications from the bell menu. This will be used _for now_ for Chat notifications _(which currently don't appear there anyway)_

### DIFF
--- a/src/hooks/useConcreteNotifications.ts
+++ b/src/hooks/useConcreteNotifications.ts
@@ -1,7 +1,7 @@
 import { useQuery } from '@apollo/client';
 import { useEffect, useState } from 'react';
 import { gql } from '../gql';
-import { Concrete_Notification } from '../gql/graphql';
+import { Concrete_Notification, Notification_Channel_Enum } from '../gql/graphql';
 import { isMessageValid } from '../helper/notification-helper';
 
 const userNotificationQuery = gql(`
@@ -17,6 +17,9 @@ const userNotificationQuery = gql(`
                     navigateTo
                 }
                 sentAt
+                notification {
+                    disabledChannels
+                }
             }
         }
     }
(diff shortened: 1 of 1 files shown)


Jonasdoubleyou (changes requested):
Should be done in backend?

JeangelLF:
> Should be done in backend?

Hmm, I also thought about that, but having that logic in the backend felt weird. 

After reading your comment, _(and thinking for a while about the reason)_ I realized we have pagination... And now it makes sense.

With this approach on the Frontend, the current app behavior doesn't change, but the current app behavior is also wrong because [this filter here](https://github.com/corona-school/user-app/blob/44627e808edfc7227a1016a7367423b4a194ef24/src/hooks/useConcreteNotifications.ts#L32) is kind of defeating the purpose of pagination...

Alright, I'll move it to the backend. Quick question @Jonasdoubleyou:

Is there a reason why the [concrete_notification](https://github.com/corona-school/backend/blob/6921540a7328f5dba7b609663309a7197b8b45a0/prisma/schema.prisma#L143) model doesn't have a relationship with the `notification` model?

JeangelLF:
> It does have a relationship through the notificationId?

No, I was referring to a proper relationship via `@relation(fields: [notificationID], references: [id])` that creates a foreign key and allows prisma to use relation queries.

For example, `message_translations` which is also cached, has a [relationship](https://github.com/corona-school/backend/blob/6921540a7328f5dba7b609663309a7197b8b45a0/prisma/schema.prisma#L949) with notification.

I thought we had another reason for this. So, if it's ok, I'll add the relationship since it would very be useful for filtering using Prisma now that we're doing the pagination with it.

Jonasdoubleyou:
It adds the additional overhead of enforcing the foreign key constraint. But as notification is rarely updated + small, this should not add too much overhead - so feel free to add an index :)

PR #598, opened 2024-07-07
language switcher and default language

MartinEichinger:
Modal Layout: 
- update Theme.ts with new variant "outlinemiddle" to cover specific button theme requirements
- update SwitchLanguageModal.tsx to change appearance of buttons

Get Selected Language:
- introduce helper getLanguageSelection.ts to use a) language from local-storage b) language set in browser c) default language (set to 'en')
- I18n.ts: load language files dependent on result of getLanguageSelection

Move Help to Settings
- introduce new column in Settings.tsx with link to 'hilfebereich'

Hamburger Menu
- update SettingsButton with new icon_menu

Language Switcher instead of Help Navigation
- update all affected files to replace HelpNavigation with new LangNavigation (language switcher)

resolves corona-school/project-user#1007


src/I18n.ts
```diff
@@ -14,18 +26,52 @@ const lazyLanguages = {
...
+export const languageList = [
+    { short: 'de', name: 'Deutsch' },
+    { short: 'en', name: 'English' },
+    { short: 'ar', name: 'اللغة العربية' },
+    { short: 'tr', name: 'Türkçe' },
+    { short: 'uk', name: 'Українська' },
+    { short: 'ru', name: 'Русский' },
+];
+
+export const languageComponents = {
```
Jonasdoubleyou, line 38: (languageIcons?)

src/I18n.ts
```diff
@@ -1,7 +1,19 @@
 import i18next from 'i18next';
 import { initReactI18next } from 'react-i18next';
+import { getLanguageSelection } from './helper/getLanguageSelection';
 
 import de from './lang/de.json';
+import en from './lang/en.json';
+import ar from './lang/ar.json';
+import uk from './lang/uk.json';
+import tr from './lang/tr.json';
+import ru from './lang/ru.json';
```
Jonasdoubleyou, line 10: Any reason to not lazy load these anymore as was done before?

src/I18n.ts
```diff
@@ -51,4 +97,11 @@ if (LANGUAGE_SWITCHER_ACTIVE) {
     }
 }
 
+// check for pre-selected language coming from lern-fair.de
+const referrer = document.referrer
+    .split(/[:/.]/)
+    .filter((elm) => elm)
+    .slice(0, -1);
+languageList.map((abbr) => referrer.includes(abbr.short) && switchLanguage(abbr.short));
```
Jonasdoubleyou, line 105: Why map? 

```
const [subdomain, domain] = document.referrer.split(".");
if (domain === "lern-fair") {
   const lng = languageList.find(...);
   if (lng) switchLanguage(lng);
}
```

src/I18n.ts
```diff
@@ -14,18 +26,52 @@ const lazyLanguages = {
...
+    ru: IconRU,
+};
+
+export const defaultLang = 'en';
+
 export const resources = {
     de: {
         translation: de,
     },
+    en: {
```
Jonasdoubleyou, line 53: not needed, see switchLanguage below

src/components/LangNavigation.tsx
```diff
@@ -0,0 +1,24 @@
+import { Button } from 'native-base';
+import React, { useState } from 'react';
+import { SwitchLanguageModal } from '../modals/SwitchLanguageModal';
+import { languageComponents } from '../I18n';
+import { getLanguageSelection } from '../helper/getLanguageSelection';
+
+const LangNavigation: React.FC = () => {
```
Jonasdoubleyou, line 7: (SwitchLanguageButton?)

src/components/LangNavigation.tsx
```diff
@@ -0,0 +1,24 @@
...
+    const [showSwitchLanguage, setShowSwitchLanguage] = useState(false);
+
+    const lang = getLanguageSelection();
```
Jonasdoubleyou, line 10: (might not reflect updates - John recently introduced useLocalStorage, maybe that could be used here too?)

src/helper/getLanguageSelection.ts
```diff
@@ -0,0 +1,17 @@
+import { languageList } from '../I18n';
```
Jonasdoubleyou, line 1: Don't think this needs to be a separate file, could be in src/I18n.ts - distributing logic across files makes it harder to follow (as we do some things on demand vs. on load)

src/pages/Appointments.tsx
```diff
@@ -125,7 +124,7 @@ const Appointments: React.FC = () => {
                 headerLeft={
                     userType !== 'screener' && (
                         <Stack alignItems="center" direction="row">
-                            <HelpNavigation />
```
Jonasdoubleyou, line 128: (so Help should go? just double checking)

Jonasdoubleyou (commented):
Looks good overall, some minor things - and please restore lazy loading of language files by removing the static import

src/I18n.ts
```diff
@@ -14,18 +26,52 @@ const lazyLanguages = {
...
+    ar: {
+        translation: ar,
+    },
 } as const;
 
 i18next.use(initReactI18next).init({
     debug: false && process.env.NODE_ENV === 'development',
     resources,
-    // The app is shipped and opened in german by default, further languages are loaded on demand
-    lng: 'de',
```
Jonasdoubleyou, line 27: This was on purpose - The initial app is in german - if users switch to a different language, that language is lazily loaded and then switched. 
Thus for the majority of users we only load the german language file.

src/I18n.ts
```diff
@@ -51,4 +97,11 @@ if (LANGUAGE_SWITCHER_ACTIVE) {
     }
```
Jonasdoubleyou, line 97: else (take over language from referrer ...) - the logic below should only be done if the user has not already set a preference

Jonasdoubleyou:
To put the lazy loading into perspective, this is how a main build looks like:

```
File sizes after gzip:

  472.08 kB  build/static/js/main.0d987cc2.js
  264.55 kB  build/static/js/556.0e41178c.chunk.js
  119.58 kB  build/static/js/735.cde0b36a.chunk.js
  110.9 kB   build/static/js/198.0631ac64.chunk.js
  43.54 kB   build/static/js/182.72c9c0d7.chunk.js
  42.79 kB   build/static/js/994.d2530f4c.chunk.js
  37.71 kB   build/static/js/933.baecd3ea.chunk.js
  30.44 kB   build/static/js/568.29b29500.chunk.js
```

And this is how this change looks like:

```
File sizes after gzip:

  1.04 MB (+563.73 kB)  build/static/js/main.7d217f8a.js
  264.55 kB             build/static/js/556.0e41178c.chunk.js
  119.37 kB             build/static/js/792.a666f68f.chunk.js
  111.23 kB (+335 B)    build/static/js/198.be65df8e.chunk.js
  24.17 kB              build/static/js/566.43786117.chunk.js
  16.93 kB              build/static/js/313.0ed2de68.chunk.js
  1.32 kB               build/static/js/29.19072f02.chunk.js
  1.13 kB               build/static/css/main.a91e9cae.css
  381 B                 build/static/css/792.31472567.chunk.css
```

As the main bundle is the only one loaded for the landing page on initial page load, for users that open the App the first time (after a new deployment) on a slow network (e.g. 3G) this would more than double the initial load from 10s (which is already bad) to more than 20s (so likely users will give up opening the app).

Not sure if this is just the import of the language files or also the rather large language icon SVGs. Either way, please try to get the bundle size (can be checked with `npm run build`) back to normal :)

src/I18n.ts
MartinEichinger: Ok. I restore the initial implementation.

src/I18n.ts
MartinEichinger: Ok. Understood

src/I18n.ts
MartinEichinger: I use it in "LangNavigation.tsx" to generate dynamically the right language icon (as component):
const Icon = languageComponents[lang as keyof typeof languageComponents];

src/I18n.ts
MartinEichinger: Thanks for your suggestion. I will take that.
(Regarding your question: I like the idea to check the subdomain and calling switchLanguage in one line (therefore map))

src/components/LangNavigation.tsx
MartinEichinger: Yes, good idea.

src/pages/Appointments.tsx
MartinEichinger: Yes. According to the issue description the link to the "hilfebereich" shall be moved to the settings page.

src/I18n.ts
MartinEichinger: Initially I thought it is a good idea to pre-set the language according to the language setting of the browser. Not understanding the lazy loading correctly, I implemented the language switch acc. to browser setting here.

Now I have understood, that the initial language setting is 'de' or according to the subdomain of the referrer.

src/I18n.ts
Jonasdoubleyou: Right, but "component" is a super generic name, "Icons" is what it is ;)

MartinEichinger: Ah, ok. Understood.

src/I18n.ts
MartinEichinger: I assume following priority, which language should be taken
1) (Initial load) Language set on lern-fair.de
1) language set in localStorage (set by user-app user)
2) default language

src/helper/getLanguageSelection.ts
MartinEichinger: I agree. At the initial load I ensure, that the localStorage('lernfair-language') is set. Then I only get the localStorage value and don't need to use that function.

src/I18n.ts
MartinEichinger: My implementation now is:
1) At the initial load I check, whether localStorage has a 'lernfair-language'. In that case I set that language again (and in case load the language file again). I assume that situation, if I reload the page
2) If not localStorage entry is available, I check whether a document.refferer with a appropriate subdomain is available and in case set this language
3) If nothing applies, I load the defaultLanguage

After that in any case a localStorage entry is available.

src/components/LangNavigation.tsx
MartinEichinger: I think it doesn't fit, as I have to set a default value.

src/I18n.ts
```diff
@@ -43,11 +68,25 @@ export async function switchLanguage(language: string) {
...
+// 3) else, according to defaultLang
+const localStorageLanguage = localStorage.getItem('lernfair-language');
+if (localStorageLanguage) {
+    /* no await, switch lazily */
+    switchLanguage(localStorageLanguage);
+} else {
+    // check for pre-selected language coming from lern-fair.de
+    const [subdomain, domain] = document.referrer.split('.');
+    if (domain === 'lern-fair') {
+        var subdomainLanguage = languageList.find((langItem) => subdomain.includes(langItem.short));
```
Jonasdoubleyou, line 83: (could be const if the following if is moved into this scope)

Jonasdoubleyou (approved):
nice!

src/I18n.ts
MartinEichinger: If I move `if (subdomainLanguage) {} else { default }` into the scope of `if (domain === 'lern-fair')`, I would need an additional else ->  `if (domain === 'lern-fair') { if (subdomainLanguage) {} else { default } } else {default}`. Without that I will need an else less.

Jonasdoubleyou: Fair enough, still could declare as let in the outer scope instead of the implicit global var

MartinEichinger: Ok

PR #599, opened 2024-07-09
Feat: Update Zoom SDK

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1239

## What was done?

- Finished React 18 Upgrade _(Required for new Zoom version)_
- React 18 broke Native-base, so I upgraded it as well to a version that fixed the issues
- Upgraded Zoom to the latest version. The only relevant things I noticed _(So far)_ are:
  - Package now lives under `@zoom/meetingsdk`
  - A few methods deprecated that don't seem to be needed anymore
  - No need to load CSS dependencies in our index.html

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -189,9 +189,6 @@
             media="(device-width: 320px) and (device-height: 568px) and (-webkit-device-pixel-ratio: 2) and (orientation: landscape)"
         />
 
-        <link type="text/css" rel="stylesheet" href="https://source.zoom.us/2.14.0/css/bootstrap.css" />
-        <link type="text/css" rel="stylesheet" href="https://source.zoom.us/2.14.0/css/react-select.css" />
-
         <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
 
         <!-- Dynamically inject a new configuration each time the app is reloaded -->
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,7 +37,7 @@
                 "@types/react-signature-canvas": "^1.0.2",
                 "@types/ua-parser-js": "^0.7.36",
                 "@types/validator": "^13.7.14",
-                "@zoomus/websdk": "2.15.2",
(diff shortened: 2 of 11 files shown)


public/index.html
```diff
@@ -189,9 +189,6 @@
             media="(device-width: 320px) and (device-height: 568px) and (-webkit-device-pixel-ratio: 2) and (orientation: landscape)"
         />
 
-        <link type="text/css" rel="stylesheet" href="https://source.zoom.us/2.14.0/css/bootstrap.css" />
```
Jonasdoubleyou, line 192: Why can we remove these (good that they are gone though)?

Jonasdoubleyou (approved):

public/index.html
JeangelLF: Thanks for highlighting this! I forgot to bring up the discussion about this.

I tried to follow the install instructions from the [new npm page](https://www.npmjs.com/package/@zoom/meetingsdk) and it doesn't mention these dependencies, however, [the official documentation](https://marketplacefront.zoom.us/sdk/meeting/web/index.html) still does.

I also saw that [since 2019](https://devforum.zoom.us/t/jquery-bootstrap-dependencies/5946) they are actively trying to get rid of the good old Bootstrap + JQuery combo

I did several tests with and without these and honestly couldn't find a single difference between both...

After testing a couple of features while doing some coverage tests, it seems they use a whopping 3.7% of the full bootstrap experience 🙈 and 0% of react-select. 
Bildschirmfoto 2024-07-10 um 14 15 08

On the old version, they're using a 7.9% for a couple of menus that are now fixed in this new version. So, with all of this, I decided to get rid of both dependencies since it's not worth loading that on all pages for a 3.7% usage.

Let me know if you have any concerns about this, maybe there are some other historical reasons for this.

Jonasdoubleyou: Then lets remove it. Rather had the opposite problem that their global CSS rules overruled some of ours ...

PR #601, opened 2024-07-11
Update interface and lists of known us from in screening app 1254

cr-eative-dev:
### What was done:

- Added a dropdown menu for the KnowsCoronaSchoolFrom sections of the student screen card (modal) and pupil screening cards and an optional text input field that becomes visible if the "Sonstiges" category is selected
- Added useEffect to automatically switch to "Sonstiges" category and display the input field with the custom KnowsFrom answer in case the Pupil has already answered that question (Shows error if over 60 characters)
- Added checks to database mutations to either save the customKnowsFrom or the standard knowsFrom that comes from the suggestions
- Removed old imports and code for obsolete TextInputWithSuggestions component
- Added "Tafel" and "Jugendzentrum" as an option to select in the pupil and student screening cards

### Closes Ticket:
https://github.com/corona-school/project-user/issues/1254


src/components/TextInputWithoutSuggestions.tsx
```diff
@@ -0,0 +1,19 @@
+import { VStack, Input } from 'native-base';
```
JeangelLF, line 1: I think we don't need this component since it's just an `Input` imported from native-base :)
So, we can replace the usage of
```jsx
<TextInputWithoutSuggestions ... />
```
with:
```jsx
<Input ... />
```
and it should probably be the same

src/widgets/screening/ScreenPupilCard.tsx
```diff
@@ -106,27 +110,43 @@ function EditScreening({ pupil, screening }: { pupil: PupilForScreening; screeni
...
     const handleOnKnowsFromChanges = (value: string) => {
-        if (knowsFromSuggestions.includes(value.trim())) {
-            setKnowsFrom(value.replaceAll('Sonstiges: ', ''));
-        } else if (value) {
-            setKnowsFrom(value.includes('Sonstiges: ') ? value : `Sonstiges: ${value}`);
+        setKnowsFrom(value);
+    };
+
+    const handleOnCustomKnowsFromChanges = (value: string) => {
+        if (value.length > 60) {
```
JeangelLF, line 134: Nice!

src/widgets/screening/ScreenPupilCard.tsx
```diff
@@ -106,27 +110,43 @@ function EditScreening({ pupil, screening }: { pupil: PupilForScreening; screeni
...
                 status: PupilScreeningStatus.Dispute,
-                knowsFrom,
+                knowsFrom: knowsFromValue,
             },
         });
         setScreeningComment(resultComment);
     }
 
     const handleOnKnowsFromChanges = (value: string) => {
-        if (knowsFromSuggestions.includes(value.trim())) {
```
JeangelLF, line 121: Let's try to keep this `Sonstiges: value` behavior since that was the expected behavior when we added this input on pupil screenings. See [here for more details](https://github.com/corona-school/project-user/issues/1179)

Did you have any trouble keeping this behavior?

JeangelLF (commented):
Thanks a lot for your PR!!

It mostly looks good! I just left some minor comments :) let me know if you need help.

src/components/TextInputWithoutSuggestions.tsx
cr-eative-dev: I removed the custom component and used the standard react native input now

src/widgets/screening/ScreenPupilCard.tsx
cr-eative-dev: I added back this string method logic for sonstiges back into the pupil screening card and the student screening card (modal)

JeangelLF: I just tested it and is still not working. I think it's because this logic now doesn't belong to this method but to the new `handleOnCustomKnowsFromChanges` which is the input onChange handler. Sorry for not clarifying.

cr-eative-dev: ok I will take a look

cr-eative-dev: @JeangelLF  not sure if I understand the purpose of this but right now I added code that adds "Sonstiges: " in front of the custom answer - is that the desired functionality?

JeangelLF (approved):
It looks great, Thank you!

PR #603, opened 2024-07-13
Update AlertMessage to use blue icon for informational content

MartinEichinger:
AlertMessage: set status of <Alert /> to "info". So the colorScheme for "info" is also used for the <Alert.Icon />

Resolves corona-school/project-user#1269

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,7 +129,7 @@
         "dev-install": "npm ci --include=optional",
         "dev-storybook": "storybook dev -p 6006",
         "dev-storybook-build": "storybook build",
-        "prettier": "./prettier-commits.sh",
+        "prettier": "prettier-commits.sh",
         "prettier-all": "prettier --write \"./**.tsx\"",
         "linter": "eslint . --ext .tsx",
         "graphql-fetch-schema": "node scripts/fetch-graphql-schema.js",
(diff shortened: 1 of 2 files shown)


src/widgets/AlertMessage.tsx
```diff
@@ -14,9 +14,9 @@ const AlertMessage: React.FC<Props> = ({ content }) => {
 
     return (
         <>
-            <Alert alignItems="start" marginY={space['0.5']} width="100%" maxWidth={backgroundWidth} colorScheme="info">
+            <Alert alignItems="start" marginY={space['0.5']} width="100%" maxWidth={backgroundWidth} status="info">
                 <HStack space={2} flexShrink={1} alignItems="center">
-                    <Alert.Icon color="danger.100" />
```
Jonasdoubleyou, line 19: (maybe we should rename the component from AlertMessage to InfoMessage, to better reflect how it looks?)

package.json
```diff
@@ -129,7 +129,7 @@
         "dev-install": "npm ci --include=optional",
         "dev-storybook": "storybook dev -p 6006",
         "dev-storybook-build": "storybook build",
-        "prettier": "./prettier-commits.sh",
+        "prettier": "prettier-commits.sh",
```
Jonasdoubleyou, line 132: (please avoid unrelated changes - but fine if we keep this in here)

Jonasdoubleyou (approved):

PR #613, opened 2024-07-17
Feat: school during registration

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1261

## What was done?

- Added some utilities to search schools making use of the [JedeSchule API](https://codefor.de/projekte/jedeschule-2/)
- If for some reason, the API fails, we should be notified via Datadog and users can still manually enter the name and continue the registration process as usual.
- Added a new optional step to collect user school
- Added a feature flag for this new step


src/config.ts
```diff
@@ -54,3 +54,4 @@ export const PROMOTE_APP_BANNER_ACTIVE = (window.liveConfig?.RUNTIME_PROMOTE_APP
 
 export const RESULT_CACHE_ACTIVE = (window.liveConfig?.RUNTIME_RESULT_CACHE_ACTIVE ?? 'false') === 'true';
 export const SERVICE_WORKER_ACTIVE = (window.liveConfig?.RUNTIME_SERVICE_WORKER_ACTIVE ?? 'false') === 'true';
+export const SCHOOL_SEARCH_ACTIVE = (window.liveConfig?.RUNTIME_SERVICE_WORKER_ACTIVE ?? 'false') === 'true';
```
Jonasdoubleyou, line 57: copy & pasta

src/lib/Schools.ts
```diff
@@ -0,0 +1,94 @@
+import { useEffect, useState } from 'react';
+import useDebounce from '../hooks/useDebounce';
+import { log, logError } from '../log';
+import { states } from '../types/lernfair/State';
+import { schooltypes } from '../types/lernfair/SchoolType';
+import { SchoolType, State } from '../gql/graphql';
+
+const API = 'https://jedeschule.codefor.de';
```
Jonasdoubleyou, line 8: Making external API requests is a privacy problem. Would prefer to make a backend request instead so that we do not share our users IP and browser info with the external service or replicate their data in our database. The later especially makes sense if we doubt that this API exists long term.

src/config.ts
JeangelLF: 🤦

src/lib/Schools.ts
Jonasdoubleyou: Ja well if you keep it in the frontend you need to discuss with legal regarding sharing user data with that API.

JeangelLF: Ah, nvm.

JeangelLF: Alrighty, [here](https://github.com/corona-school/backend/pull/1122) is it

src/config.ts
```diff
@@ -54,3 +54,4 @@ export const PROMOTE_APP_BANNER_ACTIVE = (window.liveConfig?.RUNTIME_PROMOTE_APP
 
 export const RESULT_CACHE_ACTIVE = (window.liveConfig?.RUNTIME_RESULT_CACHE_ACTIVE ?? 'false') === 'true';
 export const SERVICE_WORKER_ACTIVE = (window.liveConfig?.RUNTIME_SERVICE_WORKER_ACTIVE ?? 'false') === 'true';
+export const SCHOOL_SEARCH_ACTIVE = (window.liveConfig?.RUN_TIME_SCHOOL_SEARCH_ACTIVE ?? 'false') === 'true';
```
Jonasdoubleyou, line 57: prefix must be RUNTIME_ (https://github.com/corona-school/user-app/blob/main/serve.js#L38)

Jonasdoubleyou (commented):
otherwise good to go :)

Jonasdoubleyou (dismissed):

PR #617, opened 2024-07-19
harmonize all card sizes for topic selection

MartinEichinger:
SubjectSelectorGrouped.tsx: set card sizes to fixed height=110px and width=120px. Adjust height of heading and text breakpoint dependend, so that cards are always on same level at each viewport width.

Resolves corona-school/project-user#1278


JeangelLF (approved):
Thank you for your PR! It would be great if we could center their content in the box now that they're bigger :)

Bildschirmfoto 2024-07-22 um 09 10 53

JeangelLF (approved):
Nice 💯, thank you!

PR #622, opened 2024-07-30
Feat: Add native-base replacement

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1209

## What was done?

This is meant to be the initial step in removing `native-base` / `react-native`-related stuff from our project and, move to a more web approach now that we decided to build a PWA.

- Started building the foundations of our theme (colors/fonts/etc.) and some parts of our design system.
I made use of [TailwindCSS](https://tailwindcss.com/) for styling and [RadixUI](https://www.radix-ui.com/primitives) for our primitives _(aka. Unstyled components that provides us with basic functionality and a11y)_
- Replaced `native-base` references from the `SingleCourseStudent` page. Now we're using our own files there.
Bildschirmfoto 2024-07-31 um 12 09 40

## Next steps

My plan with this initiative is to:

- Provide a fully configured theme with all necessary colors, fonts, and customizations together with basic building blocks _(design system)_.
- Use our own files as much as possible to simplify future upgrades and component replacements.
- Execute a step-by-step migration. We **don’t need** to do it all at once. We can use the new components for new interfaces or replace native-base components with our own when updating existing functionalities.
- We are also planning to do a UI Facelift for some pages that don't have the best UI/UX, this definitely a good chance to migrate those pages as well. _(I'm looking at you, [CreateCourse](https://github.com/corona-school/project-user/issues/879) page 👀)_


JeangelLF:
@Jonasdoubleyou Before moving forward with this I wanted to get an initial review, I hope it's ok :) So I can change things as needed.

src/components/atoms/Button/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+        },
+        defaultVariants: {
+            variant: 'default',
+            size: 'default',
+        },
+    }
+);
+
+export interface ButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement>, VariantProps<typeof buttonVariants> {
+    asChild?: boolean;
```
Jonasdoubleyou, line 37: What does this parameter do?

src/pages/student/SingleCourseStudent.tsx
```diff
@@ -23,21 +19,26 @@ import { Appointment } from '../../types/lernfair/Appointment';
 import SwitchLanguageButton from '../../components/SwitchLanguageButton';
 import AppointmentsEmptyState from '../../widgets/AppointmentsEmptyState';
 import { SubcourseParticipant } from '../../types/lernfair/Course';
-import RemoveParticipantFromCourseModal from '../../modals/RemoveParticipantFromCourseModal';
+import { Button } from '@/components/atoms/Button';
+import { toast } from 'sonner';
+import RemoveParticipantFromCourseModal from '@/components/organisms/modals/RemoveParticipantFromCourseModal';
+import CancelSubCourseModal from '@/components/organisms/modals/CancelSubCourseModal';
```
Jonasdoubleyou, line 25: So in the end everything will be in the /component/ folder? Not saying that our current structure is better, but keeping folders as is makes it easier to navigate during the migration + keeps the git diff view

src/web/scss/index.scss
```diff
@@ -17,7 +17,7 @@ body {
     // max-height: -moz-available;
 
     * {
-        font-family: 'Montserrat', sans-serif !important;
+        font-family: 'Outfit Variable', sans-serif !important;
```
Jonasdoubleyou, line 20: How is this loaded?

config/webpack.config.js
```diff
@@ -300,7 +302,8 @@ module.exports = function (webpackEnv) {
           'react-dom$': 'react-dom/profiling',
           'scheduler/tracing': 'scheduler/tracing-profiling'
         }),
-        ...(modules.webpackAliases || {})
+        ...(modules.webpackAliases || {}),
+        '@': path.resolve(__dirname, '../src'),
```
Jonasdoubleyou, line 306: Do all tools understand this?

Jonasdoubleyou (dismissed):
Smaller changes please.

Also to make the migration successful, I would consider doing the following:
- Adapt the Readme
- Onboard the team (share how things shall be done in the future / some knowledge bytes, also request reviews from new team members) - i.e. I never worked with Tailwind and probably also need some onboarding (but after all it is "just CSS", so I guess everybody can work with it somehow)
- Quickly rework the main pages and then slowly expand to the "outer rim" 

Also I think we should get an overview/track progress by creating issues for every page / component. Currently not sure how much effort this is after all.

src/components/atoms/Button/index.tsx
JeangelLF: With this we can re-use our button styles to render other tags, the most common example is replacing the `button` tag with an `a`.
```jsx
    <Button asChild>
      <a href="https://lern-fair.de">Mehr erfahren</a>
    </Button>
```
This will just render an anchor with the button styles

src/pages/student/SingleCourseStudent.tsx
JeangelLF: > keeping folders as is makes it easier to navigate during the migration + keeps the git diff view

Hm, I don't want to overwhelm us with a drastic change in the folder structure 😟 But I also have to say that our current folder structure is not easy to navigate.
This is more or less how it looks right now in terms of components
```
src
|
+-- components              # All kinds of components
|   |
|   +-- achievements        # Achievements related components
    +-- notification        # Notification related components
    +-- appointments        # Some Appointments related components
    Button.tsx
    Button.stories.tsx
    Typography.tsx
    Typography.stories.tsx
    LanguageTag.tsx
    LanguageTag.stories.tsx
|
+-- modals                  # Some modals go here, some on components, some on widgets 🙈 
|
+-- widgets                 # More components of all kinds go here
|   |
    +-- course              # Some Course related components
    +-- screening           # Some Screening related components
    +-- matching            # Some Matching related components
    AlertMessage.tsx
    AlertMessage.stories.tsx
    AppointmentCard.tsx
    AppointmentCard.stories.tsx
+-- pages
|   |                   
|   +-- page-name           # Some page related-components go here, some on widgets/components

```

I'd at least suggest to take the following baby steps here to try to improve a bit:
- Unify widgets and components
- Keep the pattern of grouping feature-related components inside the `components` folder
- At the root of the `components` folder we keep only those that are meant to be shared/used in multiple features

Does that make sense?

------- 
Another thing to mention here is that I tried to implement absolute paths for components for two reasons:
1. Avoid the dreaded `../../../../`
2. Provide tooling for automatic imports so it gives priority to our `@/components` instead of components from `some-third-party-library`

src/web/scss/index.scss
JeangelLF: Via fontsource, it already contains all the `.ttf` files and the `@font-face` calls: https://github.com/corona-school/user-app/blob/bbb8a54412b9d7c3ddf28b23c78b005f37a3372c/package.json#L11

https://github.com/corona-school/user-app/blob/59813e935dc865d34069af7cf14c7848af5b576f/src/index.tsx#L8

If we prefer to do it ourselves as it was I can happily change it :)

src/components/atoms/Button/index.tsx
Jonasdoubleyou: hm, slightly confusing for me ... But if this is a common pattern lets do this, guess I need to adapt :)

src/pages/student/SingleCourseStudent.tsx
Jonasdoubleyou: Sounds good, but multiple PRs then please. The git diff is much better when not changing the folder structure at the same time.

src/web/scss/index.scss
Jonasdoubleyou: No its fine, would just double check that this is properly cached (as serve.js does some unusual things)

config/webpack.config.js
JeangelLF: Sorry, I missed this one. Yes, our tooling already understands this. TS needed a minor configuration and I also tweaked the auto-import feature to give priority to our `@/` paths over `native-base`, `react-native`, etc...

JeangelLF:
Alright, in general, I'll deliver smaller PRs :) I guess this one was the exception because of the needed configuration + adding some of the new base components.

Jonasdoubleyou (approved):

cr-eative-dev (dismissed):

PR #623, opened 2024-08-05
feat: Navigate back to course detail page after edit

cr-eative-dev:
### What was done:

- The finishCourseCreation function now accepts two parameters: errors and courseId.
- Instead of navigating back to the group page after the user edits their course, they will now automatically return to the details page of that specific course.
- The courseId value is passed to the finishCourseCreation function at the necessary points within the code.

### Closes Ticket:
https://github.com/corona-school/project-user/issues/1047


Jonasdoubleyou (approved):
We use https://www.conventionalcommits.org/en/v1.0.0/ - So the proper PR title would be "feat: Navigate back to course detail page after edit"

PR #624, opened 2024-08-07
fix: Make Match Request Card Responsive

cr-eative-dev:
### What was done:

- Enhanced the styling of "Tag.tsx" and "SubjectList.tsx" components to prevent the subject list from overflowing on the match request card when viewed on smaller screens or when many subjects are selected.

### Closes Ticket:
https://github.com/corona-school/project-user/issues/1038


JeangelLF (approved):
Nice, thank you! 🚀

PR #631, opened 2024-08-19
fix: Notifications don't execute click actions on body text

cr-eative-dev:
### What was done:

- Added the Pressable component inside the Tooltip component wrapping the tooltip-text to ensure the tooltip body text is also clickable

### Closes Ticket:
https://github.com/corona-school/project-user/issues/1284


JeangelLF (approved):
🚀

PR #634, opened 2024-08-20
Ci: Add gql codegen plugin

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1273

## What was done?

- Added the recommended `BabelOptimizerPlugin` as suggested in the [documentation](https://the-guild.dev/graphql/codegen/plugins/presets/preset-client#reducing-bundle-size)

With this change, I noticed that the big map in `gql.ts` is not included in the build anymore but instead uses the generated types.

_Note: After adding our new components the final build size increased _(as expected)_, this will unfortunately be like this until we remove native-base + friends which right now is around 22% of the build._

### DIFF
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -1,0 +1,6 @@
+const { babelOptimizerPlugin } = require('@graphql-codegen/client-preset')
+ 
+module.exports = {
+  presets: ['react-app'],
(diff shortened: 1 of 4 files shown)


Jonasdoubleyou (approved):
nice!

PR #641, opened 2024-09-04
Feat: Session manager

realmayus:
Closes https://github.com/corona-school/project-user/issues/1243.
Related backend PR: https://github.com/corona-school/backend/pull/1133

- Added session manager view accessible through settings
- Created `deviceId` which is persisted on the device 
- Added `SessionCard` component
- Adjust "user agent" of token usage

Screenshot 2024-10-21 at 11 03 58


src/components/SessionCard.tsx
```diff
@@ -0,0 +1,53 @@
...
+        icon = <IconDeviceMobile size={32} />;
+    }
+    if (deviceType === 'tablet') {
+        icon = <IconDeviceIpadHorizontal size={32} />;
+    }
+    if (deviceType === 'desktop') {
+        icon = <IconDeviceLaptop size={32} />;
+    }
+
+    // rewrite the component at the very top (native-base) in tailwind:
```
Jonasdoubleyou, line 30: outdated? Or TODO: ?

src/hooks/useApollo.tsx
```diff
@@ -298,6 +298,21 @@ const getDeviceToken = () => STORAGE.getItem('lernfair:device-token');
...
+// ----- Device ID --------------------------
+// The device ID is a permanent identifier for a device, it is used to identify
+//  the device across sessions and to revoke all sessions created by a device
+export const getDeviceId = () => STORAGE.getItem('lernfair:device-id');
+const setDeviceId = (id: string) => STORAGE.setItem('lernfair:device-id', id);
+export const getOrCreateDeviceId = () => {
+    const existing = getDeviceId();
+    if (existing) return existing;
+
+    const deviceId = crypto.randomUUID();
```
Jonasdoubleyou, line 310: nice! Did not know it - we could also use this for the other tokens

src/pages/SessionManager.tsx
```diff
@@ -0,0 +1,96 @@
...
+                            secretId: secret.id,
+                        };
+                        // try to parse the description as JSON
+                        try {
+                            const parsed = JSON.parse(description!);
+                            device.type = parsed.type;
+                            device.description =
+                                (parsed.device?.model ?? parsed.browser?.name ?? t('sessionManager.unknownDevice')) +
+                                (parsed.os?.name ? ` ${t('sessionManager.deviceOnOs')} ` + parsed.os?.name : '');
+                        } catch (e) {
```
Jonasdoubleyou, line 75: For consistency, I would prefer to drop the existing descriptions (only for type TOKEN!) in the DB

src/pages/SessionManager.tsx
```diff
@@ -0,0 +1,96 @@
...
+import React from 'react';
+
+const SessionManager: React.FC = () => {
+    const { t } = useTranslation();
+    const toast = useToast();
+
+    const sessionQuery = useQuery(
+        gql(`
+        query session {
+            me { secrets { id type description lastUsed lastUsedDeviceId } }
```
Jonasdoubleyou, line 21: Do we filter out expired secrets (expiredAt < Date.now())?

Jonasdoubleyou (dismissed):

src/pages/SessionManager.tsx
Jonasdoubleyou: Aah, we anyways have the support access etc.

src/pages/SessionManager.tsx
realmayus: Yes, that is part of the field resolver.

src/Utility.ts
```diff
@@ -11,20 +11,25 @@ export const TOKEN_LENGTH = 32;
 export const REDIRECT_PASSWORD = `/login`;
 
 export const toTimerString = (referenceDate: DateTime, theDate: DateTime) => {
```
JeangelLF, line 13: Could we translate the texts we generate with this function?
Probably we could use luxon to do that, https://moment.github.io/luxon/api-docs/index.html#datetimetorelative

What do you think? 👀

src/pages/SessionManager.tsx
```diff
@@ -0,0 +1,96 @@
+import { useMutation, useQuery } from '@apollo/client';
+import { Spacer, Stack, useToast } from 'native-base';
```
JeangelLF, line 2: Could we use some of the new components here?

For the `useToast` you could use `import { toast } from 'sonner';` _(Which btw I need to move to a separate file that belong to us)_

JeangelLF (commented):
It's looking good! 😃

src/pages/SessionManager.tsx
realmayus: It's for the icons as well as for translating the "on" in "Chrome on Android"

src/Utility.ts
realmayus: Good idea, done (also created a follow-up issue: https://github.com/corona-school/project-user/issues/1357)

Jonasdoubleyou (approved):

PR #642, opened 2024-09-05
fix: Skip cache for appointment queries

Jonasdoubleyou:
Unfortunately pagination in GraphQL is terribly broken, and we update the query result of an existing query with fetchMore(...) passing in other variables, without changing the original variables of the query. Thus this messes up the cache.

Unfortunately for some reason cachePolicy: "no-cache" does not work and the cache is still taken :/ Thus adding a hack to skip the cache, by adding NO_CACHE in the query name

https://github.com/corona-school/project-user/issues/1310

### DIFF
--- a/src/hooks/useApollo.tsx
+++ b/src/hooks/useApollo.tsx
@@ -72,6 +72,10 @@ class FullResultCache extends ApolloCache<NormalizedCacheObject> {
         }
 
         const name = definition.name?.value;
+        if (name?.includes('NO_CACHE')) {
+            return undefined;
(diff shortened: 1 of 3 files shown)


Fabian5150 (approved):
Both PRs look good to me, which approach do you want to choose?

Jonasdoubleyou:
Both maybe? :) - I anyways disliked the fetchMore logic so refactoring it out seemed good anyways. And having an option to completely disable the cache for certain queries also seems like a good idea ...

PR #643, opened 2024-09-05
Feat/screener course edit

Fabian5150:
Closes part of https://github.com/corona-school/project-user/issues/1201
(Appointments can NOT be edited, the rest of the course can now
@FForstbach decided that the rest of the ticket is not needed for now as the need for this feature is not that big)


src/pages/Appointment.tsx
```diff
@@ -93,20 +127,38 @@ const Appointment: React.FC<AppointmentParams> = ({ startMeeting }) => {
...
+        data: screenerAppointment,
+        loading: isLoadingScreenerAppointment,
+        error: screenerAppointmentError,
+    } = useQuery(COURSE_SCREENER_APPOINTMENT_VIEW, { variables: { appointmentId }, skip: userType !== 'screener' });
+
+    const data = studentAppointment ?? pupilAppointment ?? screenerAppointment;
+    const loading = isLoadingStudentAppointment ?? isLoadingpupilAppointment ?? isLoadingScreenerAppointment;
+    const error = studentAppointmentError ?? pupilAppointmentError ?? screenerAppointmentError;
+
+    // Redirect course screeners if they landed on an appointment page that is not a group appointment
```
Jonasdoubleyou, line 155: Would prefer if we prohibit this in the backend instead of "hiding" in the frontend ... Also if they "landed on a wrong appointment" I guess it is better to show an error then to hide the problem by redirecting to /start

src/pages/CreateCourse.tsx
```diff
@@ -125,7 +127,12 @@ const CreateCourse: React.FC = () => {
     const [currentIndex, setCurrentIndex] = useState<number>(state?.currentStep ? state.currentStep : 0);
     const isEditing = useMemo(() => !!prefillCourseId, [prefillCourseId]);
 
-    const { data: studentData, loading } = useQuery(
+    const [loadingStudent, setLoadingStudent] = useState(userType === 'student');
+    const [studentMyself, setStudentMyself] = useState<{ firstname: string; lastname: string }>();
+    const [canCreateCourse, setCanCreateCourse] = useState<{ allowed: boolean; reason: string }>();
+    const [studentId, setStudentId] = useState<Number>();
```
Jonasdoubleyou, line 133: (number would be the more correct type, Number is a number object - which is something very exotic ;))

src/pages/CreateCourse.tsx
```diff
@@ -125,7 +127,12 @@ const CreateCourse: React.FC = () => {
     const [currentIndex, setCurrentIndex] = useState<number>(state?.currentStep ? state.currentStep : 0);
     const isEditing = useMemo(() => !!prefillCourseId, [prefillCourseId]);
 
-    const { data: studentData, loading } = useQuery(
+    const [loadingStudent, setLoadingStudent] = useState(userType === 'student');
+    const [studentMyself, setStudentMyself] = useState<{ firstname: string; lastname: string }>();
```
Jonasdoubleyou, line 131: Maybe worth a comment that these are not set for screeners

src/pages/screening/SingleCourseScreener.tsx
```diff
@@ -85,6 +85,11 @@ const SingleCourseScreener: React.FC = () => {
     const navigate = useNavigate();
     const { space, sizes } = useTheme();
 
+    const location = useLocation();
+    const locState = location?.state as {
+        wasEdited: boolean;
```
Jonasdoubleyou, line 90: Where is this location state set?

src/routing/NavigatorLazy.tsx
```diff
@@ -363,7 +363,7 @@ export default function NavigatorLazy() {
                 path="/edit-appointment/:id"
                 element={
                     <RequireAuth>
-                        <SwitchUserType pupilComponent={<Dashboard />} studentComponent={<EditAppointment />} />
+                        <SwitchUserType pupilComponent={<Dashboard />} studentComponent={<EditAppointment />} screenerComponent={<EditAppointment />} />
```
Jonasdoubleyou, line 366: While at it, could switch here to <RequireRole> .... the <Dashboard> fallback for pupils makes little sense

Jonasdoubleyou (commented):
Looks good overall, some minor stuff :)

src/pages/screening/SingleCourseScreener.tsx
Fabian5150: in CreateCourse, when the course editing has been finished. (in the "finishCourseCreation"-Callback)

Fabian5150:
As mentioned in the backend PR I removed the screeners access to the anything regarding appointments as this feature is currently not needed and should not be further developed for now.

Jonasdoubleyou (approved):

PR #661, opened 2024-09-17
Feat: Info modal for iOS

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1338

## What was done?

- Updated old instructions component with new info modal

Bildschirmfoto 2024-09-17 um 09 44 58

### DIFF
--- a/src/lang/ar.json
+++ b/src/lang/ar.json
@@ -2387,8 +2387,8 @@
         "iOSInstallInstructions": {
             "title": "تثبيت التطبيق",
             "steps": {
-                "first": "Klicke auf <0/> oben oder unten im Browser",
-                "second": "Wähle <0/> \"Zum Home-Bildschirm\""
+                "first": "انقر على <0/> في أعلى أو أسفل المتصفح",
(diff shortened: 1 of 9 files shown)


src/lang/ar.json
```diff
@@ -2384,6 +2383,13 @@
             "install": "تثبيت التطبيق",
             "not-supported": "لسوء الحظ، فإن تثبيت التطبيق على هذا الجهاز غير مدعوم.",
             "installed": "التطبيق مثبت بالفعل."
+        },
+        "iOSInstallInstructions": {
+            "title": "تثبيت التطبيق",
+            "steps": {
+                "first": "Klicke auf <0/> oben oder unten im Browser",
```
realmayus, line 2390: We should translate these strings (all languages have missing translations)

realmayus (changes requested):

src/lang/ar.json
JeangelLF: Good catch! I think the translation failed because of the `<0/>`.

realmayus (approved):

PR #681, opened 2024-10-15
fix: Cache watchers behavior

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1353

## What was done?

- Move watchers to a separate map to avoid losing them when the cache entry is removed. With this, features like `refetchQueries` on `useMutations` work again _(Which fixes the reported issue)_

### DIFF
--- a/src/hooks/useApollo.tsx
+++ b/src/hooks/useApollo.tsx
@@ -43,10 +43,11 @@ const LOGIN_WITH_DEVICE_TOKEN_MUTATION = gql(`
     }
   `);
 
+type Watcher = () => void;
+
 interface FullResult {
     data: any;
     variables: string;
-    watchers: (() => void)[];
 }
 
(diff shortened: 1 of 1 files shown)


src/hooks/useApollo.tsx
```diff
@@ -149,19 +151,24 @@ class FullResultCache extends ApolloCache<NormalizedCacheObject> {
...
 
         const watcher = () => {
             log('GraphQL Cache', 'fire watcher');
             watch.callback(this.diff(watch));
         };
 
-        if (entry) {
+        if (name) {
+            log('GraphQL Cache', 'store watcher');
+            this.watchers.set(name, this.watchers.get(name)?.concat(watcher) ?? []);
```
Jonasdoubleyou, line 163: If name but not entry you might still want to return the cleanup function?

Jonasdoubleyou (dismissed):

PR #682, opened 2024-10-15
Add QR code to login with authtoken-link

MartinEichinger:
Added qr-code to AppointmentDetail page, when zoom-link is active.

Added new graphql query (ShortTimeAccessHelper) to create a short time login token.

Add qrCodeLink to AppointmentMetaDetail and the qr-code badge according to the proposed design.
Related text is added to de.json and translated accordingly.

Added new package to package.json -> qrcode.react.


MartinEichinger:
Änderungen sollten soweit gemacht sein. Aktuell fehlt noch die Implementierung bzgl. matomo -> Hinweise?

package.json
```diff
@@ -228,7 +228,8 @@
     },
     "dependencies": {
         "compression": "^1.7.4",
-        "express": "^4.18.2"
+        "express": "^4.18.2",
+        "qrcode.react": "^4.0.1"
```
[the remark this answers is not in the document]
MartinEichinger: done

src/components/appointment/AppointmentDetail.tsx
```diff
@@ -37,6 +37,7 @@ const AppointmentDetail: React.FC<AppointmentDetailProps> = ({ appointment }) =>
     const [canceled, setCanceled] = useState<boolean>(false);
     const [showDeleteModal, setShowDeleteModal] = useState<boolean>(false);
     const [showDeclineModal, setShowDeclineModal] = useState<boolean>(false);
+    const [loginHref, setLoginHref] = useState<string>('empty');
```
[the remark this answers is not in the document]
MartinEichinger: done

src/components/appointment/AppointmentDetail.tsx
```diff
@@ -110,9 +114,32 @@ const AppointmentDetail: React.FC<AppointmentDetailProps> = ({ appointment }) =>
 
     const isLastAppointment = useMemo(
         () => (appointment.appointmentType === Lecture_Appointmenttype_Enum.Group && appointment.total === 1 ? true : false),
+        // eslint-disable-next-line react-hooks/exhaustive-deps
         [appointment.total]
     );
 
+    useEffect(() => {
+        createShortTimeLoginData();
```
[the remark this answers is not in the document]
MartinEichinger: done

src/components/appointment/AppointmentMetaDetails.tsx
```diff
@@ -67,80 +76,121 @@ const AppointmentMetaDetails: React.FC<MetaProps> = ({
     const isAppointmentOver = useMemo(() => {
         const end = DateTime.fromISO(startDateTime).plus({ minutes: duration + 15 });
         return end < DateTime.now();
+        // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
 
+    const canJoin = null;
+    const canStartMeeting = useMemo(
```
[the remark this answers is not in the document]
Jonasdoubleyou: helper function sounds good, we should avoid duplicating this rather complex logic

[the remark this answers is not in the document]
MartinEichinger: Introduced custom hook useCanJoinMeeting. canJoinMeeting was used before in different ways e.g. start as optional value or as required value. Therefore the custom hook is a bit more effordeous.

src/components/VideoButton.tsx
```diff
@@ -71,10 +71,7 @@ const VideoButton: React.FC<VideoButtonProps> = ({
         }
     };
 
-    const canStartMeeting = useMemo(
-        () => canJoin ?? (startDateTime && duration && canJoinMeeting(startDateTime, duration, isInstructor ? 240 : 10, DateTime.now())),
-        [canJoin, duration, isInstructor, startDateTime]
-    );
+    const canStartMeeting = useCanJoinMeeting(startDateTime!, duration!, isInstructor ? 240 : 10, DateTime.now(), canJoin);
```
Jonasdoubleyou, line 74: Here this changed from canJoin ?? (...) to canJoin || (...) - was this deliberate?

IMO a parameter "canJoin" to a function "useCanJoin" is a bit strange - Maybe rather do the condition here:

 = canJoin ?? canJoinMeeting(...)

src/hooks/useCanJoinMeeting.tsx
```diff
@@ -0,0 +1,23 @@
+import { DateTime } from 'luxon';
+import { useMemo } from 'react';
+
+export const useCanJoinMeeting = (start: string, duration: number, joinBeforeMinutes: number, now: DateTime, canJoin?: boolean) => {
```
Jonasdoubleyou, line 4: Do we refresh all users every N seconds to update the "now" parameter? (I think we had this mechanism somewhere). Maybe it makes sense to move now from a parameter into the function as a state, and then update it using useInterval(...) to consistently add the time update to all users?

src/widgets/AppointmentCard.tsx
```diff
@@ -159,7 +159,7 @@ const AppointmentCard: React.FC<Props> = ({
         return maxParticipants - participantsCount;
     }, [maxParticipants, participantsCount]);
 
-    const isCurrent = _dateNext && duration ? canJoinMeeting(_dateNext, duration, isOrganizer ? 240 : 10, DateTime.now()) : false;
+    const isCurrent = useCanJoinMeeting(_dateNext!, duration!, isOrganizer ? 240 : 10, DateTime.now());
```
Jonasdoubleyou, line 162: Are _dateNext and duration always set? If yes, maybe make them required in the component params?

src/components/VideoButton.tsx
MartinEichinger: Agree. I will use a const canStartMeeting to use useCanJoinMeeting (w/o canJoin). I will then use canJoin with an ?? operator and canStart Meeting.

src/hooks/useCanJoinMeeting.tsx
MartinEichinger: I change the parameter now to a state, which I update every minute by using useInterval.

src/widgets/AppointmentCard.tsx
MartinEichinger: It is not set always. Setting _datenext and duration as required param leads to errors in some of the calling components.

Jonasdoubleyou: Right but then the not null assertion is wrong?

MartinEichinger: Ok

src/components/appointment/AppointmentMetaDetails.tsx
```diff
@@ -67,80 +77,140 @@ const AppointmentMetaDetails: React.FC<MetaProps> = ({
...
+    );
+
+    const createShortTimeLoginData = async () => {
+        const expiresAt = DateTime.now().plus({ hours: 1 });
+        const res = await createShortTimeLoginToken({ variables: { expiresAt: expiresAt, description: `` } });
+        const token = res?.data?.tokenCreate;
+
+        setLoginURL(
+            process.env.NODE_ENV === 'production'
+                ? `https://app.lern-fair.de/login-token?secret_token=${token}&temporary`
```
Jonasdoubleyou, line 103: Couldn't you use https://{window.location.host}/login-token? That would also work for staging + review apps

src/components/appointment/AppointmentMetaDetails.tsx
```diff
@@ -67,80 +77,140 @@ const AppointmentMetaDetails: React.FC<MetaProps> = ({
...
+                                <IconPointFilled size="6" />
+                                <Text fontSize="xs">{t('appointment.detail.qrcode.bp3')}</Text>
+                            </HStack>
+                            <HStack space={2} alignItems={'center'} mt={2}>
+                                <Text fontSize="xs">{t('appointment.detail.qrcode.footer')}</Text>
+                                <IconArrowNarrowRight size="24" stroke={2} />
+                            </HStack>
+                        </VStack>
+                        <VStack>
+                            <QRCodeSVG value={loginURL} />
```
Jonasdoubleyou, line 209: If loginURL is still loading, what do we show?

Jonasdoubleyou (dismissed):
Looks good, some final comments :)

MartinEichinger:
@Jonasdoubleyou 
> Looks good, some final comments :)

Currently the implementation regarding matomo is missing. Unfortunately I don't understand what to do here. Any hints?

src/components/appointment/AppointmentMetaDetails.tsx
MartinEichinger: Ok. But I still have to differ as the localhost URL starts with 'http' (instead of 'https').

src/components/appointment/AppointmentMetaDetails.tsx
MartinEichinger: We show a QR code with content 'empty'.

src/components/appointment/AppointmentMetaDetails.tsx
Jonasdoubleyou: https://developer.mozilla.org/en-US/docs/Web/API/Location/origin maybe then .origin instead?

Jonasdoubleyou:
Hi Martin, thought it was easier to just draft something for the Matomo thing than to explain it - https://github.com/corona-school/user-app/pull/703 :) - Does that make sense to you?

Jonasdoubleyou (approved):
good to go after rebase :)

src/components/appointment/AppointmentMetaDetails.tsx
MartinEichinger: Many thanks for the hint. Passt.

Jonasdoubleyou (dismissed):

src/components/appointment/AppointmentMetaDetails.tsx
```diff
@@ -67,80 +77,136 @@ const AppointmentMetaDetails: React.FC<MetaProps> = ({
...
+            mutation LoginTokenForSecondDevice($expiresAt: DateTime!, $description: String!) { tokenCreate(expiresAt: $expiresAt, description: $description) }
+        `)
+    );
+
+    const createShortTimeLoginData = async () => {
+        const expiresAt = DateTime.now().plus({ hours: 1 });
+        const res = await createShortTimeLoginToken({ variables: { expiresAt: expiresAt, description: `` } });
+        const token = res?.data?.tokenCreate;
+
+        setLoginURL(`${window.location.origin}/login-token?secret_token=${token}&temporary`);
```
Jonasdoubleyou, line 101: Ah, maybe remove &temporary and add referrer=qr_code

Jonasdoubleyou (dismissed):
no more comments - everything is awesome :)

PR #685, opened 2024-10-16
fix missing line wrap for federal state icon

MartinEichinger:
Introduce forced line wrap for federal state icon in student and pupil UpdateData.

Resolves corona-school/project-user#1351


JeangelLF (approved):
Looks good, thank you!

PR #693, opened 2024-10-26
SuS: update selectable subjects

MartinEichinger:
For SuS same selectable subjects shall be shown as for HuH.
Updated SubjectSelector.tsx with more specific subject arrays SUBJECTS_MAIN, SUBJECTS_MINOR and SUBJECTS_RARE (which are used for HuH).

Resolves corona-school/project-user#1350


JeangelLF (approved):
NIce!

PR #702, opened 2024-11-02
Add section certificate at matching

MartinEichinger:
Add section certificate at matching. CTA goes to "/profile". Used text added in language files.

Resolves https://github.com/corona-school/project-user/issues/1319


src/pages/student/MatchingStudent.tsx
```diff
@@ -233,13 +233,24 @@ const MatchingStudent: React.FC<Props> = () => {
...
                                 </VStack>
                             </VStack>
                         )}
+                        <VStack space={space['0.5']} paddingX={space['1']} width="100%" marginX="auto" maxWidth={ContainerWidth}>
+                            <Heading paddingBottom={space['0.5']}>{t('matching.volunteering.title')}</Heading>
+                            <Text maxWidth={ContentContainerWidth} paddingBottom={space['0.5']}>
+                                {t('matching.volunteering.text')}
+                            </Text>
+                            <VStack marginBottom={space['2.5']}>
+                                <Button width={ButtonContainer} onPress={() => navigate('/profile')}>
```
JeangelLF, line 249: - The button variant should be `outline` _(bonus if you use the new component)_ 😅 
- Could we use an anchor to bring the user directly to the certificates section?

MartinEichinger: @JeangelLF Ok. 
1) Button is not part of the radix-ui primitives, but part of themes. I would then install @radix-ui/themes. When do we use outline and when do we use solid?
2) I tried to use 'useRef' within useEffect. As it doesn't work, I found a solution with storing the anchor element within a state and use a goto function.

JeangelLF: >Button is not part of the radix-ui primitives, but part of themes. I would then install @radix-ui/themes.

We already have a Button component that uses Radix under the hood. Take a look [here](https://github.com/corona-school/user-app/blob/main/src/pages/student/single-course/StudentCourseButtons.tsx#L50) for example, we're doing:
```tsx
import { Button } from '@/components/Button';
// usage
<Button
    onClick={() => {
        navigate('/edit-course', {
            state: { courseId: subcourse.id },
        });
    }}
    variant="outline"
    leftIcon={<IconPencil size={16} />}
    className="w-full  md:w-fit"
>
    {t('single.courseInfo.editCourse')}
</Button>
```

>When do we use outline and when do we use solid?

About when do we use what, it's still not _very_ well defined, but in general, we try to use `outline` for actions that are not that important. The green solid version is for actions that may cause a change in something  "Save changes of X thing", and the yellow version should ideally be a very important action we want the user to take, probably "Create course" or something like that. Unfortunately, the existing UI doesn't follow this pattern very well but we'd like to move forward in that direction :)

>I tried to use 'useRef' within useEffect. As it doesn't work, I found a solution with storing the anchor element within a state and use a goto function.

Probably works for now 💯, In the future we'll redesign a bit the Settings page

MartinEichinger: Regarding the button and radix-ui; I have the impression, that the buttons need to be changed on every occursion. Also the customization of the theme needs to be done than. Might it be better to solve this within an own issue?

MartinEichinger: Ok

JeangelLF (approved):
Nice 💯!

PR #705, opened 2024-11-06
Update label kurse und max matches message

FForstbach:


realmayus (changes requested):
We should translate the changes to the other languages

PR #708, opened 2024-11-11
feat: Sus huh screening show error message if language is not selected

cr-eative-dev:
### What was done:

- Added responsive vertical layout to SuS + HuH Screening to improve scanability
- Added error messages to SuS + HuH Screening to make sure language, grade and subjects are always selected

### Closes Ticket:
https://github.com/corona-school/project-user/issues/1297


src/widgets/screening/ScreenStudentCard.tsx
```diff
@@ -246,14 +263,14 @@ export function ScreenStudentCard({ student, refresh }: { student: StudentForScr
...
                 {t('helper')} / {student.firstname} {student.lastname}
             </Heading>
-            <HStack>
+            <VStack space={space['2']}>
                 <LanguageTagList languages={student.languages} />
-                <Text fontSize="20px" lineHeight="50px">
-                    {' '}
-                    -{' '}
-                </Text>
+                {languageError && <Text color={colors.error[500]}>{languageError}</Text>}
```
JeangelLF, line 268: The errors here are a bit weird because Screeners cannot edit these fields for HuH.
Do you think this is something we could include? :) It should be similar to how it works on the `ScreenPupilCard` component with the Update Language feature.

The query should be something like: 
```ts
const UPDATE_LANGUAGES_QUERY = gql(`
    mutation StudentUpdateLanguages($studentId: Float!, $languages: [StudentLanguage!]) { studentUpdate(studentId: $studentId, data: { languages: $languages }) }
`);
```

cr-eative-dev: ok I will add those buttons and queries - wasnt sure if I should

JeangelLF (approved):

PR #719, opened 2024-11-16
Update english lang flag

MartinEichinger:
Introduce new library flag-icons. 

Update language icon flags with flag-icons icons. Updates done in I18n.ts, SwitchLanguageModal.tsx and SwitchLanguageButton.tsx.

Change flag for english language.

Resolves corona-school/project-user#1371

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
         "eslint-plugin-lernfair-app-linter": "./linter/",
         "eslint-plugin-unused-imports": "^2.0.0",
         "eslint-webpack-plugin": "^3.1.1",
+        "flag-icons": "^7.2.3",
         "fs-extra": "^10.0.0",
         "graphql": "^16.5.0",
         "html-webpack-plugin": "^5.5.0",
--- a/public/index.html
+++ b/public/index.html
@@ -188,6 +188,7 @@
             href="/splash/apple-splash-1136-640.jpg"
             media="(device-width: 320px) and (device-height: 568px) and (-webkit-device-pixel-ratio: 2) and (orientation: landscape)"
         />
+        <link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/lipis/flag-icons@7.2.3/css/flag-icons.min.css" />
 
         <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
 
@@ -199,7 +200,7 @@
 
     <body>
         <noscript>You need to enable JavaScript to run this app.</noscript>
-        <div id="root" style="width: 100vw; min-height: 100dvh; overflow-x: hidden;"></div>
+        <div id="root" style="width: 100vw; min-height: 100dvh; overflow-x: hidden"></div>
         <div id="zmmtg-root" class="dd-privacy-hidden" style="display: none"></div>
         <!-- <div id="aria-notify-area"></div> -->
     </body>
--- a/src/I18n.ts
+++ b/src/I18n.ts
@@ -2,12 +2,6 @@ import i18next from 'i18next';
 import { initReactI18next } from 'react-i18next';
 
 import de from './lang/de.json';
-import IconDE from './assets/icons/icon_flag_de.svg';
-import IconEN from './assets/icons/icon_flag_en.svg';
-import IconUK from './assets/icons/icon_flag_uk.svg';
-import IconTR from './assets/icons/icon_flag_tr.svg';
-import IconRU from './assets/icons/icon_flag_ru.svg';
-import IconAR from './assets/icons/icon_flag_ar.svg';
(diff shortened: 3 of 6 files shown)


public/index.html
```diff
@@ -188,6 +188,7 @@
             href="/splash/apple-splash-1136-640.jpg"
             media="(device-width: 320px) and (device-height: 568px) and (-webkit-device-pixel-ratio: 2) and (orientation: landscape)"
         />
+        <link rel="stylesheet" href="https://cdn.jsdelivr.net/gh/lipis/flag-icons@7.2.3/css/flag-icons.min.css" />
```
JeangelLF, line 191: Wasn't this library working without the CDN? 👀 I saw this CSS could be imported from the installed package.
If that wasn't working I'd suggest a different library (e.g https://www.npmjs.com/package/country-flag-icons which seems more actively maintained)

src/I18n.ts
```diff
@@ -2,12 +2,6 @@ import i18next from 'i18next';
 import { initReactI18next } from 'react-i18next';
 
 import de from './lang/de.json';
-import IconDE from './assets/icons/icon_flag_de.svg';
```
JeangelLF, line 5: If we're going with a library we should remove these svg files

src/components/SwitchLanguageButton.tsx
```diff
@@ -23,7 +23,7 @@ const SwitchLanguageButton = ({ className }: SwitchLanguageButtonProps) => {
                 variant="none"
                 size="icon"
             >
-                <Icon />
+                <span className={`fi ${Icon} bg-center bg-cover rounded-full h-7 w-7`}></span>
```
JeangelLF, line 26: Just a minor detail, I think a size of 5 or 6 would be nicer for the flags so they don't look that big 😅

JeangelLF (changes requested):

public/index.html
MartinEichinger: I tried to import the CSS directly in SwitchLanguageButton.tsx and SwitchLanguageModal.tsx, but till now it didn't work. Also tried some variations of the import.
With country-flag-icons I have the problem, that so far I could not manage to change the size of the flag so that it fully covers the rounded button.

src/I18n.ts
MartinEichinger: Do you mean to remove the svg files from the assets folder?

JeangelLF: Yep, unless those files are being used elsewhere I don't think we have to keep them around.

public/index.html
JeangelLF: > I tried to import the CSS directly in SwitchLanguageButton.tsx and SwitchLanguageModal.tsx, but till now it didn't work. Also tried some variations of the import.

Yeah, I saw some issues about that on their GitHub page. Let's move on with the other package, we may have better luck 😅.

> With country-flag-icons I have the problem, that so far I could not manage to change the size of the flag so that it fully covers the rounded button.

I see. Could you push your current changes? I could then provide some insights about what could be missing 🙂

MartinEichinger: Ok

src/I18n.ts
MartinEichinger: Ok

public/index.html
MartinEichinger: I updated the file SwitchLanguageButton.tsx (changes are only for testing purpose). This is the button on the upper right corner with the rounded flag.

src/components/SwitchLanguageButton.tsx
```diff
@@ -3,11 +3,16 @@ import { SwitchLanguageModal } from '../modals/SwitchLanguageModal';
 import { languageIcons } from '../I18n';
 import { Button } from './Button';
 import { cn } from '@/lib/Tailwind';
+//import '/node_modules/flag-icons/css/flag-icons.min.css';
+import { DE } from 'country-flag-icons/react/3x2';
```
JeangelLF, line 7: This should do the trick:

```suggestion
import { DE } from 'country-flag-icons/react/1x1';
```

package.json
```diff
@@ -63,6 +64,7 @@
         "eslint-plugin-lernfair-app-linter": "./linter/",
         "eslint-plugin-unused-imports": "^2.0.0",
         "eslint-webpack-plugin": "^3.1.1",
+        "flag-icons": "^7.2.3",
```
JeangelLF, line 67: The last one, I swear 😅.
Let's uninstall this library since we're not using it anymore

MartinEichinger: Ah, I have overseen it. Ok.

JeangelLF (approved):
Very nice, thank you! 💯

PR #729, opened 2024-11-26
fix: change icons of lang cards

MartinEichinger:
Update icons of language cards with country flags.
Component IconLoader can now be called with the path of the icon or with the icon name, which is defined in I18n.ts/languageList

Change is implemented in the profile pages for the students and the pupils.

Resolves corona-school/project-user#1372


src/I18n.ts
```diff
@@ -14,22 +14,59 @@ const lazyLanguages = {
...
+    { short: 'al', long: 'albanisch', name: '' },
+    { short: 'am', long: 'armenisch', name: '' },
+    { short: 'ar', long: 'arabisch', name: 'اللغة العربية' },
+    { short: 'az', long: 'aserbaidschanisch', name: '' },
+    { short: 'ba', long: 'bosnisch', name: '' },
+    { short: 'bg', long: 'bulgarisch', name: '' },
+    { short: 'cn', long: 'chinesisch', name: '' },
+    { short: 'de', long: 'deutsch', name: 'Deutsch' },
+    { short: 'en', long: 'englisch', name: 'English' },
+    { short: 'es', long: 'spanisch', name: 'Espana' },
```
JeangelLF, line 36: Do we use this `name` property? If so, why some of them don't have it 👀 ?

JeangelLF (approved):
Looks good!, I just left one non-blocking question.

src/I18n.ts
MartinEichinger: You are right. I just structured it according to languageListSelectionModal, but here it isn't needed. I will delete it.

MartinEichinger: Done

JeangelLF (approved):
Nice!

PR #733, opened 2024-11-30
fix: hide info if already screened

MartinEichinger:
Updated logic to create a screening link, so that screeing link is not created/sent, if student has already a tutor or instructor role:

(stud.canCreateCourse= 'not-screened' && stud.canReqMatch = 'not-screened') ||
            (stud.canCreateCourse = 'not-instructor' && stud.canRequestMatch = 'not-screened') ||
            (stud.canCreateCourse = 'not-screened' && stud.canReqMatch = 'not-tutor') ||
            (stud.canCreateCourse = 'not-instructor' && stud.canReqMatch = 'not-tutor')

Resolves corona-school/project-user#1042

### DIFF
--- a/src/widgets/ImportantInformation.tsx
+++ b/src/widgets/ImportantInformation.tsx
@@ -191,8 +191,9 @@ const ImportantInformation: React.FC<Props> = ({ variant }) => {
 
         // -------- Screening -----------
         if (
-            student?.canRequestMatch?.reason === 'not-screened' ||
-            student?.canCreateCourse?.reason === 'not-screened' ||
+            (student?.canCreateCourse?.reason === 'not-screened' && student?.canRequestMatch?.reason === 'not-screened') ||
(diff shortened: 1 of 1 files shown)


src/widgets/ImportantInformation.tsx
```diff
@@ -191,8 +191,9 @@ const ImportantInformation: React.FC<Props> = ({ variant }) => {
 
         // -------- Screening -----------
         if (
-            student?.canRequestMatch?.reason === 'not-screened' ||
-            student?.canCreateCourse?.reason === 'not-screened' ||
+            (student?.canCreateCourse?.reason === 'not-screened' && student?.canRequestMatch?.reason === 'not-screened') ||
```
Jonasdoubleyou, line 194: I think we could simplify the logic if we look at the roles:

const wasScreenedBefore = roles.includes('TUTOR') || roles.includes('INSTRUCTOR');

if (!wasScreenedBefore && (...))

Jonasdoubleyou (approved):
The issue is quite old, maybe you could quickly double check with Support (Paula) whether we still dont screen on a second role.

PR #744, opened 2025-01-02
fix: PushManager does not exist on older devices

Jonasdoubleyou:
Accessing sw.pushManager.(some method) failed


JeangelLF (approved):
Thank you!

PR #745, opened 2025-01-05
fix: only show courses in either past or current section

dhenkel92:
fixes https://github.com/corona-school/project-user/issues/1405


dhenkel92:
It still relies on [this logic](https://github.com/corona-school/backend/blob/master/common/courses/filters.ts#L19-L43), which says a course is in the past if every lecture is over since at least a day.

PR #751, opened 2025-01-10
Feat: Allow email change after registration

JeangelLF:
## Ticket

Resolves https://github.com/corona-school/project-user/issues/1406

## What was done?

- Allow users to change their email address after registration. This is the same behavior that users have on their settings page. The only difference is that the form is presented in a modal.

Bildschirmfoto 2025-01-10 um 15 11 08


src/lang/de.json
```diff
@@ -491,12 +493,15 @@
             "label": "Helfer:in"
         },
         "verifyemail": {
-            "title": "Überprüfe dein E-Mail-Postfach!",
-            "mailsendto": "Wir haben eine E-Mail an {{email}} gesendet.",
-            "description": "Bevor du unser Angebot nutzen kannst, musst du deine E-Mail-Adresse bestätigen. Danach wirst du automatisch in deinen User-Bereich weitergeleitet.",
-            "notreceived": "Keine E-Mail erhalten? Schau bitte auch im Spam nach!",
+            "title": "Bestätige deine Email-Adresse",
```
Jonasdoubleyou, line 496: (Fun fact - the old writing before "E-Mail" was the correct one (https://www.duden.de/rechtschreibung/E_Mail), "Email" is actually an old german word for something else related to pottery (https://www.duden.de/rechtschreibung/Email) ... But nowadays many people write it wrongly, so ...)

Jonasdoubleyou (approved):

src/lang/de.json
JeangelLF: Interesting 🤓. Thanks a lot for the feedback!
I've just updated it to be correct again.

Jonasdoubleyou (approved):

PR #760, opened 2025-01-20
Feat: add free text field for other dissolve reason

realmayus:
Add free text field for other dissolve reason

~~TODO: is it required for users to enter a free text value?~~

Closes https://github.com/corona-school/project-user/issues/1403


realmayus:
I implemented autofocus and rewrote the modal using tailwind. I'm not sure how to address the problem of cut-off content on small screens though... maybe we should modify the Modal component to have a max height of `100vh` and have a scrollable content container?

src/modals/DissolveMatchModal.tsx
```diff
@@ -38,65 +42,86 @@ const DissolveMatchModal: React.FC<DissolveModalProps> = ({ showDissolveModal, a
     ];
 
     return (
-        <Modal isOpen={showDissolveModal} onClose={onPressBack}>
-            <Modal.Content>
-                <Modal.CloseButton />
+        <Modal
+            isOpen={!!showDissolveModal}
+            onOpenChange={(isOpen) => isOpen && onPressBack()}
```
JeangelLF, line 47: I think this should be `!isOpen && onPressBack()`
Currently the modal can't be closed with the `X` 😅

src/modals/DissolveMatchModal.tsx
```diff
@@ -38,65 +42,86 @@ const DissolveMatchModal: React.FC<DissolveModalProps> = ({ showDissolveModal, a
...
+                                    placeholder={t('matching.dissolve.modal.otherFreeText')}
+                                    value={otherFreeText}
+                                    onChange={(e) => setOtherFreeText(e.target.value)}
+                                    className="w-full my-2"
+                                    autoFocus={true}
+                                />
+                            )}
+                        </div>
+                        <ModalFooter variant="default">
+                            <Button onClick={onPressBack} variant="ghost">
```
JeangelLF, line 111: Let's add the following classes to the buttons so they look a bit better on mobile:
```tsx
className="w-full lg:w-fit"
```

src/modals/DissolveMatchModal.tsx
realmayus: Thanks!

JeangelLF (approved):
Looks good 🚀 Thank you!

PR #762, opened 2025-01-21
Fix: change window.open to '_blank'

MartinEichinger:
Zoom Meeting Modal. Open new window in new tab.

Resolves corona-school/project-user#1389


JeangelLF (commented):
Could we also update it on the [AdhocMeetingModal](https://github.com/corona-school/user-app/blob/3d274236cabc0455beb8df8ba9ec947f5c108168/src/modals/AdHocMeetingModal.tsx#L64)? :)

JeangelLF (approved):
Nicee!

PR #773, opened 2025-02-08
fix: achievement modal size on safari

dhenkel92:
I'm not sure if there is any other impact, but it solves the problem. Fixes: https://github.com/corona-school/project-user/issues/1408

Before:
before

After:
after


dhenkel92:
This change is part of the following stack:

- #773 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

Jonasdoubleyou (approved):

PR #776, opened 2025-02-10
Feat: Screening tool adjustments

JeangelLF:
## Ticket

Resolves (partially) https://github.com/corona-school/project-user/issues/1425

## What was done?

- Added a new tab for drafts
- Each course state is in a separate tab
- Show the first lecture date/instructor in the course card _(only for screeners)_
- Increased the course limits from 10 to 50
- Remove `SuS / HuH-Empfehlung` prefix when showing suggestions
- Add a save button to only save the pupil screening comment


src/pages/screening/ScreenerGroup.tsx
```diff
@@ -48,10 +48,28 @@ const ScreenerGroup: React.FC = () => {
...
+            }
+            instructors {
+                firstname
+                lastname
+            }
             }
         }  
     `);
 
+    const {
```
Jonasdoubleyou, line 62: Might be easier to query all courses at once, then .filter on the state for the tabs?

src/widgets/AppointmentCard.tsx
```diff
@@ -335,6 +342,23 @@ const AppointmentCard: React.FC<Props> = ({
...
+                                    <div className="flex flex-col gap-y-2">
+                                        {firstLecture ? (
+                                            <Typography>
+                                                <span className="font-bold">Kursstart:</span> {formatDate(firstLecture?.start)}
+                                            </Typography>
+                                        ) : null}
+                                        <Typography>
+                                            {instructors && instructors?.length ? (
+                                                <span>
+                                                    <span className="font-bold">Kursleiter:</span> {instructors[0].firstname} {instructors[0].lastname}
```
Jonasdoubleyou, line 355: Why just the first one? (Okay, should be the 95% case)

Jonasdoubleyou (dismissed):

src/widgets/AppointmentCard.tsx
Jonasdoubleyou: Ah, probably too less space

src/pages/screening/ScreenerGroup.tsx
Jonasdoubleyou: Right, or turn courseState into a state and refetch when switching tabs?

Alternatively you could also do multiple courseSearch in the same query (with a GraphQL alias)

Jonasdoubleyou (approved):

PR #822, opened 2025-05-16
fix: Naming & usage of courseId/subcourseId on the CreateCourse page

JeangelLF:
## What was done?

- Fixed the naming & usage of the courseId and subcourseId on the CreateCourse page. It seems we were using them interchangeably, and it was causing issues on Production _(since there, they are not the same, unlike on STG where they happened to be the same for all courses)_

### DIFF
--- a/src/pages/CreateCourse.tsx
+++ b/src/pages/CreateCourse.tsx
@@ -97,8 +97,8 @@ const CreateCourse: React.FC = () => {
     const toast = useToast();
 
     const location = useLocation();
-    const state = location.state as { courseId?: number; currentStep?: number };
-    const prefillCourseId = state?.courseId;
+    const state = location.state as { subcourseId?: number; currentStep?: number };
+    const prefillSubcourseId = state?.subcourseId;
     const breadcrumbRoutes = useBreadcrumbRoutes();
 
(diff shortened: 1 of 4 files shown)


Jonasdoubleyou (approved):
This is just renaming, right? No actual change to the logic?

Jonasdoubleyou:
> It seems we were using them interchangeably, and it was causing issues on Production 

yes this did hold for a long time as we did not have a mechanism to add further subcourses to a course, and then this wrong assumption started to creep in ...

JeangelLF:
> This is just renaming, right? No actual change to the logic?

99% of it was renaming _(mainly the `prefillCourseId` ~> `prefillSubcourseId`)_. The only two parts that were using the wrong variable were: [Here](https://github.com/corona-school/user-app/pull/822/files#diff-e7fd2b7bc26cc586d466f61393361d1e0c5b0ddbbc3c4a1a5b0a7cf00168ccc4L562) and [here](https://github.com/corona-school/user-app/pull/822/files#diff-e7fd2b7bc26cc586d466f61393361d1e0c5b0ddbbc3c4a1a5b0a7cf00168ccc4R610).
My bad, next time I'll make it in two PRs or at least commits

PR #847, opened 2025-07-16
Feat: Login UI facelift

JeangelLF:
## Ticket

Resolves https://github.com/corona-school/project-user/issues/1504

## What was done?

- Rebuild the login page according to the new design
- Removed all uses of native-base from it


src/components/SwitchLanguageButton.tsx
```diff
@@ -1,29 +1,50 @@
...
-                className={cn('group rounded-full hover:bg-primary-light hover:brightness-105', className)}
-                onClick={() => setShowSwitchLanguage(true)}
-                variant="none"
-                size="icon"
-            >
-                <Icon className={cn(`rounded-full h-5 w-5 border`)} />
-            </Button>
-            <SwitchLanguageModal isOpen={showSwitchLanguage} onIsOpenChange={setShowSwitchLanguage} />
+            <DropdownMenu>
+                <DropdownMenuTrigger asChild>
```
Jonasdoubleyou, line 21: What will this trigger if variant === 'modal'? Wouldn't it be cleaner to have two variants, and duplicate the button?

Jonasdoubleyou (approved):
Looks beautiful!

When entering the username + password, there is no visual indication that I am on the Lern-Fair page ... Not sure if it needs one, but if I am searching for my password for some time I might be confused when coming back.

The welcome-loki.svg adds another 12KB to our initial load ;) ...

JeangelLF:
> When entering the username + password, there is no visual indication that I am on the Lern-Fair page ... Not sure if it needs one, but if I am searching for my password for some time I might be confused when coming back.

That's a great catch. I'll pass the feedback to Jessica. I think we should at least keep the logo somewhere in the navbar

> The welcome-loki.svg adds another 12KB to our initial load ;) ...

Yeah 😅. But in exchange, we'll remove the 10KB `lf-logo.svg` and the 228KB `lf-bg.png`. Not in this PR because we're also using that in the Registration flow. _(That one comes next)_

PR #851, opened 2025-08-03
Feat/zip code to state

Fabian5150:
## What was done?
- Added location field for student screeners
- Added automatic mapping from zip code to state for student screeners
- Added errors for zip codes of wrong length and zip codes with occurences in multiple states

Closes additional points of https://github.com/corona-school/project-user/issues/1311

### DIFF
--- a/src/pages/screening/student/PersonalDetails.tsx
+++ b/src/pages/screening/student/PersonalDetails.tsx
@@ -15,7 +15,7 @@ import { EditLanguagesModal } from '@/widgets/screening/EditLanguagesModal';
 import { EditSubjectsModal } from '@/widgets/screening/EditSubjectsModal';
 import { ApolloError, useMutation } from '@apollo/client';
 import { IconCheck, IconDeviceFloppy, IconKey, IconTestPipe } from '@tabler/icons-react';
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { toast } from 'sonner';
 import { ButtonField } from '../components/ButtonField';
@@ -32,7 +32,6 @@ interface PersonalDetailsProps {
 interface FormErrors {
     languages?: string;
     subjects?: string;
-    zipCode?: string;
 }
 
 const CREATE_LOGIN_TOKEN_MUTATION = gql(`
@@ -62,11 +61,20 @@ const PersonalDetails = ({ student, refresh }: PersonalDetailsProps) => {
     const [descriptionForScreening, setDescriptionForScreening] = useState(student.descriptionForScreening);
     const [location, setLocation] = useState(student.state);
     const [zipCode, setZipCode] = useState(student.zipCode ?? '');
+    const [zipCodeError, setZipCodeError] = useState('');
 
     const [errors, setErrors] = useState<FormErrors>({});
 
     const [mutationUpdateStudent, { loading: isUpdating }] = useMutation(UPDATE_STUDENT_MUTATION);
     const [mutationCreateLoginToken] = useMutation(CREATE_LOGIN_TOKEN_MUTATION);
+
+    useEffect(() => {
+        if (!location || location === Student_State_Enum.Other) return;
+
+        if (zipCode.length !== zipCodeLength()) {
+            setZipCodeError(`Postleitzahl für ${t(asTranslationKey(`lernfair.states.${location}`))} muss ${zipCodeLength()} Ziffern haben.`);
+        }
+    }, [zipCode, location]);
 
     const zipCodeLength = () => {
         switch (location) {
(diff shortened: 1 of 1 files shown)


src/pages/screening/student/PersonalDetails.tsx
```diff
@@ -56,13 +59,56 @@ const PersonalDetails = ({ student, refresh }: PersonalDetailsProps) => {
...
+    const zipCodeToState = (input: string) => {
+        if (!input || input.length !== 5) return;
+
+        // Finds all states where the zipCode occurs in
+        const states = zipStateMapping.filter((range) => range.min <= input && range.max >= input).map((range) => range.state);
+
+        if (states.length === 0) {
+            setZipCodeError('Keine gültige deutsche PLZ');
+        } else if (!states.every((state) => state === states[0])) {
+            const statesUnique = states.filter((item, index) => states.indexOf(item) === index);
```
Jonasdoubleyou, line 102: Why is this deduplication needed? Shouldnt it be unique per state?

src/pages/screening/student/PersonalDetails.tsx
```diff
@@ -56,13 +59,56 @@ const PersonalDetails = ({ student, refresh }: PersonalDetailsProps) => {
...
+            case undefined:
+                return null;
+            default:
+                return 5;
+        }
+    };
+
+    // Range check here purposefully lexicographic to allow comparisons with codes with leading zeros
+    const zipCodeToState = (input: string) => {
+        if (!input || input.length !== 5) return;
```
Jonasdoubleyou, line 94: If you enter a 5 digit zip code for location = CH, will this still set the state? 

Ah, you prevent that through the max length input ...

src/pages/screening/student/PersonalDetails.tsx
```diff
@@ -56,13 +59,56 @@ const PersonalDetails = ({ student, refresh }: PersonalDetailsProps) => {
...
+
+        // Finds all states where the zipCode occurs in
+        const states = zipStateMapping.filter((range) => range.min <= input && range.max >= input).map((range) => range.state);
+
+        if (states.length === 0) {
+            setZipCodeError('Keine gültige deutsche PLZ');
+        } else if (!states.every((state) => state === states[0])) {
+            const statesUnique = states.filter((item, index) => states.indexOf(item) === index);
+
+            setZipCodeError(`PLZ gibt es in mehreren Bundesländern: ${statesUnique.map((s) => t(asTranslationKey(`lernfair.states.${s}`))).join(', ')}`);
```
Jonasdoubleyou, line 104: What if the location was already chosen?

src/pages/screening/student/PersonalDetails.tsx
```diff
@@ -75,7 +121,8 @@ const PersonalDetails = ({ student, refresh }: PersonalDetailsProps) => {
                         gender: (gender as Gender) || undefined,
                         descriptionForMatch,
                         descriptionForScreening,
-                        zipCode,
+                        state: location as any,
+                        zipCode: zipCode !== '' ? zipCode : null,
```
Jonasdoubleyou, line 125: So this can be submitted even if there is a validation error?

Jonasdoubleyou (approved):

src/pages/screening/student/PersonalDetails.tsx
Fabian5150: I had some cases where I got the same state multiple times. 
(And I double checked, there are really multiple ranges for one state that 'catch' the same zip code)
Not quite sure if the mapping file is just not that good, or if the german zip code system is so messy, that it actually has to be that way.

src/pages/screening/student/PersonalDetails.tsx
Fabian5150: The mapping currently only contains german zip codes,
so the location has to be set manually for Austria/Switzerland/others.

src/pages/screening/student/PersonalDetails.tsx
Jonasdoubleyou: Well if there are two overlapping ranges, you can always merge into one ... But okay, I guess its also not worth it to fix the file

src/pages/screening/student/PersonalDetails.tsx
Jonasdoubleyou: Hm, what if I choose "Other" and then enter a 5 digit number that happens to collide with a german zip code? Shouldn't we early exit here if the state is already set in general?

src/pages/screening/student/PersonalDetails.tsx
Fabian5150: I had to do it this way as users currently get initialized with a state of 'other'.
So the automatic mapping would be blocked right from the beginning

src/pages/screening/student/PersonalDetails.tsx
Fabian5150: I wasn't quite sure with this one, as I wouldn't want an empty zip code or a zip code with multiple state occurences from blocking the screener.
Even with "unvalid german zip code" I'm not quite certain, if there aren't any special cases, where we would falsely detect it as unvalid.

As this is just for screeners, I wanted to prefer just having the error as a notice for them, instead of blocking them.

src/pages/screening/student/PersonalDetails.tsx
Fabian5150: I would have prefered to have a 'cleaner' file too, but I think it's not worth the effort

PR #857, opened 2025-08-12
feat: Restore scroll when navigating to the referrals page

Jonasdoubleyou:


src/hooks/useScrollRestoration.ts
```diff
@@ -67,3 +67,12 @@ export function useRestoredNumberState(initial: number, scrollGroup: string) {
 
     return [value, setValue] as const;
 }
+
+// On component mount, scroll #root (the element defined in index.html that scrolls) to the top
+// This overrides the browser behaviour of restoring the scroll position while navigating between pages.
+export function useScrollToTop() {
+    useEffect(() => {
+        const el = document.getElementById('root')!;
```
Jonasdoubleyou, line 75: the trick is to scroll #root and not the window ;)

src/pages/Referrals.tsx
```diff
@@ -56,6 +57,8 @@ const Referrals: React.FC<{}> = () => {
         }
     };
 
+    useScrollToTop();
```
Jonasdoubleyou, line 60: Not sure if we want this more globally or only on selected pages ... I totally forgot, but we also have this "restore scroll" stuff in some pages, that could collide

Fabian5150:
Screenshot_2025-08-12-21-53-33-15_3aea4af51f236e4932235fdada7d1643

On Android (Firefox Mobile) I still have the problem. :/
My initial scroll position is as seen in the picture.

Jonasdoubleyou:
Strange, works on iOS :/

Fabian5150:
Works in Chrome 👍
And now it did work again in Firefox... huh

src/pages/Referrals.tsx
Fabian5150: I think locally is fine.
I did not notice the issue much with other pages

Fabian5150:
Works in Chrome 👍
And now in Firefox too... huh

PR #858, opened 2025-08-14
Feat: Registration steps 5

JeangelLF:
## Ticket

https://github.com/corona-school/project-user/issues/1506

## What was done?

- Added the last two steps of the registration process OurRules / RegistrationCompleted
- Now all the post-meeting-booked data is being saved after each step
- Added some extra functions to quickly read/save the system/marketing notifications preferences
- Updated the Registration page. Almost all of the Registration logic now lives in `useRegistrationForm`
- Removed unused components from the previous registration (not all of them yet)
- Updated the navigation to verify-email and book a screening appointment to go to the Registration page
- Updated translations (some of these will change soon)


src/pages/registration/useRegistrationForm.tsx
```diff
@@ -37,13 +38,10 @@ export interface RegistrationForm {
         name: string;
         zip?: string | null;
         city?: string | null;
-        schooltype?: SchoolType | null;
-    };
-    zipCode: number;
-    notificationPreferences: {
-        importantInformation: boolean;
-        recommendations: boolean;
+        schooltype?: string | null;
```
Jonasdoubleyou, line 41: string? Wasn't SchoolType more accurate?

src/pages/registration/useRegistrationForm.tsx
```diff
@@ -143,6 +167,57 @@ export const RegistrationProvider = ({ children }: { children: React.ReactNode }
...
+            if (values.currentStep === RegistrationStep.grade) {
+                await updateProfile({ variables: { gradeAsInt: values.grade } });
+            } else if ([RegistrationStep.school, RegistrationStep.schoolType, RegistrationStep.zipCode].includes(values.currentStep)) {
+                await updateProfile({
+                    variables: {
+                        school: {
+                            schooltype: values.school.schooltype as unknown as SchoolType,
+                            zip: values.school.zip ?? values.zipCode.toString(),
+                            city: values.school.city,
+                            name: values.school.name,
```
Jonasdoubleyou, line 183: If we store this at each of the intermediate steps with half-filled data, wouldn't this create lots of duplicate school entries (https://github.com/corona-school/backend/blob/a026515beb9dabf034ae1b189c9d3c2ae6869608/graphql/pupil/mutations.ts#L207) ?

src/User.tsx
```diff
@@ -23,13 +21,13 @@ export const RequireAuth = ({ children, isRetainPath = true }: { children: JSX.E
 
         // Require pupils and students to be verified
         if (user && !user.screener && !(user.pupil ?? user.student)!.verifiedAt) {
-            return <VerifyEmailModal email={user.email} userType={user.pupil ? 'pupil' : 'student'} />;
+            return <Navigate to="/registration" />;
```
Jonasdoubleyou, line 24: How do these jump to the right step? Where is that handled?

src/hooks/useNotificationPreferences.ts
```diff
@@ -47,18 +53,37 @@ const useUserPreferences = () => {
...
-    const systemNotificationCategories = getSystemNotificationPreferenceCategories();
-    const currentSystemPreferences = Object.entries(userPreferences).filter(([key]) => systemNotificationCategories.includes(key));
+    const currentSystemPreferences = Object.entries(userPreferences).filter(([key]) => getSystemNotificationPreferenceCategories().includes(key));
+    const currentMarketingPreferences = Object.entries(userPreferences).filter(([key]) => getMarketingNotificationPreferenceCategories().includes(key));
 
     const hasPushSystemNotificationsEnabled = currentSystemPreferences.every(([key, value]) => {
         return value.push === true;
     });
 
+    const hasSystemPreferencesEnabled = currentSystemPreferences.every(([key, value]) => {
```
Jonasdoubleyou, line 63: "every" ... Shouldn't it be "some"? Hm, either way could cause confusion if this is "off" but one still gets marketing notification.

Jonasdoubleyou (approved):

src/pages/registration/useRegistrationForm.tsx
JeangelLF: Right, I think I changed for a moment after giving up after unsuccessfully dealing with the two SchoolTypes enums we have. I updated it to use at least one of them and then casting the other one where needed. These duplicated enums might be something I should address in the future ...

src/pages/registration/useRegistrationForm.tsx
JeangelLF: Yeah, good point. I updated it to save only after we have all the school information.

Jonasdoubleyou (approved):

PR #868, opened 2025-08-21
Feat: Registration dropdown menu

JeangelLF:
## What was done?

- Updated some modals used on the registration funnel with the new components
- Added the RegistrationDropdownMenu component to allow logout/account deactivation/notification settings/contact support
- Temp fix for modal issue after being open from a dropdown item

## Preview

Bildschirmfoto 2025-08-25 um 11 41 33
Bildschirmfoto 2025-08-25 um 11 41 57

### DIFF
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -47,10 +47,6 @@ const ModalContent = React.forwardRef<React.ElementRef<typeof DialogPrimitive.Co
                         e.preventDefault();
                     }
                 }}
-                onCloseAutoFocus={(event) => {
-                    event.preventDefault();
-                    document.body.style.pointerEvents = '';
-                }}
             >
                 {children}
                 <DialogPrimitive.Close
--- a/src/pages/registration/DropdownRegistrationMenu.tsx
+++ b/src/pages/registration/DropdownRegistrationMenu.tsx
@@ -27,7 +27,7 @@ export const DropdownRegistrationMenu = () => {
 
     return (
         <>
-            <DropdownMenu>
+            <DropdownMenu modal={false}>
                 <DropdownMenuTrigger asChild>
                     <Button className={cn('group rounded-full hover:bg-primary-light hover:brightness-105')} variant="none" size="icon">
                         <span className="rounded-full">


src/components/Modal.tsx
```diff
@@ -47,6 +47,10 @@ const ModalContent = React.forwardRef<React.ElementRef<typeof DialogPrimitive.Co
                         e.preventDefault();
                     }
                 }}
+                onCloseAutoFocus={(event) => {
+                    event.preventDefault();
+                    document.body.style.pointerEvents = '';
```
Jonasdoubleyou, line 52: What does this do? Shouldnt we set it globally in some CSS file?

src/pages/registration/DropdownRegistrationMenu.tsx
```diff
@@ -0,0 +1,69 @@
...
+                </DropdownMenuTrigger>
+                <DropdownMenuContent align="end" className="border-none rounded-xl flex flex-col gap-y-2 p-2 min-w-[235px]">
+                    <DropdownMenuArrow />
+                    {form.currentStep !== RegistrationStep.confirmEmail && (
+                        <DropdownMenuItem className="gap-x-[10px] h-12" onClick={() => setIsNotificationPreferencesOpen(true)}>
+                            <IconBell className="!size-6" />
+                            <Typography className="font-semibold leading-3">{t('registration.menu.notifications')}</Typography>
+                        </DropdownMenuItem>
+                    )}
+                    {form.currentStep !== RegistrationStep.confirmEmail && (
```
Jonasdoubleyou, line 46: Why is this hidden in the confirmEmail step?

Jonasdoubleyou (approved):
Do we have a link to /impressum reachable somewhere? IIRC it must be reachable on every page according to [german law](https://www.gesetze-im-internet.de/ddg/__5.html) ("unmittelbar erreichbar"), so maybe this could also be added to the dropdown?

src/pages/registration/DropdownRegistrationMenu.tsx
JeangelLF: This keeps more or less the same options for this step we had in the previous version.
Maybe we could allow the contact support option _(even though in this step there is already an option for that)_.
I think notification preferences doesn't make much sense since the email is not even verified at that point and account deactivation also makes sense I guess? 🤔

JeangelLF:
> Do we have a link to /impressum reachable somewhere? IIRC it must be reachable on every page according to [german law](https://www.gesetze-im-internet.de/ddg/__5.html) ("unmittelbar erreichbar"), so maybe this could also be added to the dropdown?

I shared this with Jessica to see where this can be added. I'd include it in a separate PR, but thank you for the heads up! 🙌

PR #876, opened 2025-09-08
Feat: Registration steps adjustments

JeangelLF:
## What was done?

- Added the provided portraits in the bookAppointment step
- Added validation to optionally skip some school steps if the school was found from the list.
- Added validation set a minimum step for users that already completed the all registration process.


src/assets/images/registration/portraits_huh.png
JeangelLF, line 1: Should we put these type of things in our S3 instead? 🤔  @Jonasdoubleyou

Jonasdoubleyou (approved):

PR #972, opened 2026-04-07
Feat: Videochat start adjustments

JeangelLF:
## What was done?

- Allow instructors to join a meeting 48h in advance
- Allow users to join past meetings within 30 days

### DIFF
--- a/src/components/VideoButton.tsx
+++ b/src/components/VideoButton.tsx
@@ -10,6 +10,7 @@ import { IconVideo } from '@tabler/icons-react';
 import { useCanJoinMeeting } from '@/hooks/useCanJoinMeeting';
 import { toast } from 'sonner';
 import { INSTRUCTOR_JOIN_IN_ADVANCE_MINUTES, PARTICIPANT_JOIN_IN_ADVANCE_MINUTES } from '@/Utility';
+import { DateTime } from 'luxon';
 
 type VideoButtonProps = {
     isInstructor?: boolean;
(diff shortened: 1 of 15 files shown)


src/components/VideoButton.tsx
```diff
@@ -85,7 +87,12 @@ const VideoButton: React.FC<VideoButtonProps> = ({
         }
     };
 
-    const canStartMeeting = useCanJoinMeeting(isInstructor ? 240 : 10, startDateTime, duration);
+    const canStartMeeting = useCanJoinMeeting(isInstructor ? INSTRUCTOR_JOIN_IN_ADVANCE_MINUTES : PARTICIPANT_JOIN_IN_ADVANCE_MINUTES, startDateTime, duration);
+    const isDisabled = () => {
+        // Allow users to join meetings that are already over. (only if it was in the last 30 days).
+        if (isOver && startDateTime && DateTime.fromISO(startDateTime).diffNow('days').days > -30) return false;
```
Jonasdoubleyou: move into useCanJoinMeeting?

src/components/VideoButton.tsx
```diff
@@ -85,7 +87,12 @@ const VideoButton: React.FC<VideoButtonProps> = ({
...
+        return !(canJoin ?? canStartMeeting);
```
Jonasdoubleyou: Where do we pass in canJoin? If we pass it in, we ignore your new logic?

src/components/VideoButton.tsx
Jonasdoubleyou: useCanJoinMeeting is anyways strange - maybe we should directly pass in the Appointment?

src/lang/de.json
```diff
@@ -2503,7 +2503,8 @@
         "Pupil": "Schüler:in",
         "Retiree": "Rentner:in",
         "Student": "Student:in",
-        "Azubi": "Auszubildende:r"
+        "Azubi": "Auszubildende:r",
+        "SeekingEmployment": "Arbeitssuchend"
```
Jonasdoubleyou, line 2507: a bit random ...

Jonasdoubleyou (dismissed):

src/lang/de.json
Jonasdoubleyou: Ah, related to your backend change

src/components/VideoButton.tsx
JeangelLF: We were using it in an weird way by passing through the same value returned by `useCanJoinMeeting`. I removed that usage, but kept the prop to preserve its intended purpose as a "potential override" for the internal disable logic.

Jonasdoubleyou (approved):

